PR #8, opened 2014-02-12
Update economy.cpp

sharcashmo:
In the old code the block when market type is SELL and item type is TRADE was never reached. However the intention (imho) was that both improvement is increased in both selling and buying market.

If it is not the case, it would be even better to remove the SELL & TRADE block.

Also the 

```
if (m->type == M_SELL)
```

condition has been removed, as markets are either SELL or BUY, and the condition is always fulfilled.


jt-traub:
This has been added to PR #189 and this is being closed.

PR #15, opened 2015-11-17
eliminate logically dead code

ennorehling:
Each of these changes eliminates a piece of code that can logically never run, which means they should change no behavior. I assume they are artifacts of some earlier refactorings, or the result of a misunderstanding. Shorter code is easier to read, and nobody likes to read dead code (at best it wastes time, at worst, it confuses the reader).

### DIFF
--- a/battle.cpp
+++ b/battle.cpp
@@ -444,7 +444,6 @@ void Game::GetDFacs(ARegion * r,Unit * t,AList & facs)
 						break;
 					}
 				}
-				if (AlliesIncluded == 1) break; // forlist(units)
 			}
 			if (AlliesIncluded == 1) break; // forlist (objects)
 		}
--- a/economy.cpp
+++ b/economy.cpp
@@ -1457,7 +1457,6 @@ void ARegion::Migrate()
 	int totalimm = 0;
 	forlist(&migfrom) {
 		ARegion *r = (ARegion *) elem;
-		if (!r) continue;
 		
 		// figure range
 		int xdist = r->xloc - xloc;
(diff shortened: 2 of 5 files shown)


jt-traub:
I am going to not merge these as while I am not a fan of dead code, I am slowly working my way through a bunch of things, and there are many sort of dead ends in the code.  Most of these will go away (I hope) as code quality overall improves.

ennorehling:
But how does code quality improve if you're not merging changes that try to improve it?

jt-traub:
> But how does code quality improve if you're not merging changes that try to improve it?

Fair question.  My main answer is that I fully intend to replace a string and a list usages with std::string and std:: containers over time so this code eventual dies off.   It just didn’t feel urgent to fix these 6 or so extraneous checks in the interim.  I’ve started on that work (for instance the Afile stuff has been excised) but the list and string stuff hasn’t really been tackled yet.

ennorehling:
Good answer. Custom replacements for the standard strings and containers are seldom a good idea.

jt-traub:
> Good answer. Custom replacements for the standard strings and containers are seldom a good idea.

Agreed.  I am sad I didn't do it 20 years ago when I maintained the 4.x branch for a while, but I also recognize that it is quite likely the std library wasn't as flexible/powerful as it is today and while still a bad idea, they were less of a bad idea then than they are today.  

I also fully intend to remove the custom RNG in place of the stdlib random functions and, as you might have noticed, I added a unit test framework recently as well and have been building unit tests at least for things I change...  Things improve over time and that's all you can do.  I haven't spent a lot of time looking at your Eressea codebase, but I can see you have done a lot of the things I hope to do over time, though I do hope to avoid external lib dependencies and submodules if I can :) [mostly because those are fraught for people who *aren't* deeply steeped in development experience]

PR #16, opened 2015-11-17
Use of non-array delete

ennorehling:
While most implementations do the right thing for delete on scalars, it's still a bug not to use delete[]. Especially on an array of objects (destructors need to be called), as in the case of GetNewUnit (game.cpp).

### DIFF
--- a/astring.cpp
+++ b/astring.cpp
@@ -79,7 +79,7 @@ AString::AString(char c)
 
 AString::~AString()
 {
-	if (str) delete str;
+	if (str) delete[] str;
 	str = NULL;
 }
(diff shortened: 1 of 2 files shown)


valdisz:
@artyomtrityak we could incorporate this PR into `neworigins` branch, it removes some warnings that are obviously bugs in `AString`

jt-traub:
While a number of these had previously been cleaned up a couple weren't.   The compiler wasn't warning on them, mostly because compilers have gotten better, but the code correctness is technically better.  Closing this as it's been pulled into PR #189

PR #25, opened 2016-06-26
Fix edit region terrain and then g key

snowgoon88:
When editing a region, the 'g' option (regenerate all according to terrain) => segfault
because
 - all products where deleted (products.DeleteAll() )
 - production was recreated, but only for non I_SIVER items (SetupProds() )
 - in [economy.cpp] SetupEditRegion() searched for Production of I_SILVER => null pointer.


sgb:
Thanks for the patch - it's a good catch in a part of the code that doesn't see a lot of attention!

snowgoon88:
You are welcome.

In fact, I'm trying to help our non-programmer MJ to add new levels
(Underdeep) to an existing game, so I need to get into the code. And try to
devise the correct tool for him :o)

Alain

PR #32, opened 2019-07-28
Change Transmutation

artyomtrityak:
Allow to redefine Transmutation output


artyomtrityak:
Too mant changes :( Will create separated PR with only transmutation changes

PR #35, opened 2019-08-24
Bug: Fix men on horse can tax without riding

artyomtrityak:
If unit has horses but does not have RIDING skill, it should not be possible to tax with `GameDefs::TAX_HORSE_AND_RIDING_SKILL` setting and without `GameDefs::TAX_HORSE` setting


artyomtrityak:
18+ years old :)

PR #36, opened 2019-09-24
make arcadia compile

tgerigk:
Tried to improve readability of Makefile. And made Arcadia compile.


sgb:
While I'm in favour of making arcadia work, this sadly does not compile under linux.  Do you need more details?

tgerigk:
Does it compile now? Otherwise i need more details.
Here it compiles (on Debian Stretch).

sgb:
It does compile now!

I'm unsure as to why you went to all the trouble of changing all the variants to reference "../game.h" and "../gamedata.h" instead if just leaving in the "-I.." in CXXFLAGS though.  This does mean that an arcadia specific patch affects everything else, when it doesn't really need to.

Are you intending to do more work on arcadia?  Because although it does compile, it generates far more warnings than I'm used to.

tgerigk:
I know. Fixing compiler-warnings is on my to-do-list ( <- is this a real english word or is that some kind of pseudo-anglicism? ) Would be new branch / new pull-request.

sgb:
It's a common English construction, if not an actual word; we often refer to a to-do list (note only one hyphen there).  What is your native language?

I'd prefer it if this branch could restore the "-I.." to the CXXFLAGS and revert the changes to header file locations before being merged though.

tgerigk:
My native language ist german.

Did some reading...

-  https://en.cppreference.com/w/cpp/preprocessor/include basically says search order is "implementation-defined". No guaranties here.

-  https://gcc.gnu.org/onlinedocs/gcc-9.2.0/gcc/Directory-Options.html#Directory-Options : recommends -iquote instead of -I when using #include "file". Nice explanation of search order.

- Clang supports -iquote. But i couldn't find information about search order. Can only hope that they are compatible to gcc.

Do you prefer -iquote or -I?

sgb:
I'd prefer -I, as that was what was in the Makefile's CFLAGS before the changes (I'm not sure why you removed it from the CXXFLAGS?), and I usually prefer to make the smallest set of changes that achieve a goal.

Note that the gcc.gnu.org link does not recommend -iquote over "-I", but over "-I-" : the second form has a trailing dash.  We weren't using the second form anyway, so the recommendation doesn't apply here.

Your English is very good for a non-native speaker; I wouldn't have guessed if you hadn't asked for native speaker advice!

tgerigk:
Done: changed #includes back to "file.h" instead of "where-to-find/file.h"; re-added "-I." to CXXFLAGS.

Why i removed -I. in the first place? uncertainty..
With header-files like aregion.h, astring.h, etc.h existing in both --- arcadia/ and ./ --- i thought that #include "location/header.h" is less ambiguous. Using "-I." i was unsure where the compiler will look for header-files. Or if i should add "-I." conditionally to CXXFLAGS depending on which game one is compiling.

hmm.. guess i read that somewhere else .. skimmed over a lot of texts which google gave me ..

.. most of the time i am looking up how to spell words ..

PR #37, opened 2019-10-06
Fixing compiler warnings

tgerigk:
work-in-progress / dont merge yet

most fixes applied to arcadia/ did already exist in ./

PR #38, opened 2019-10-06
fixes to compiler warnings which need review

tgerigk:
Don't want to introduce logic-errors.

Note: these fixes will be part of fix_warnings pull request after discussion.

### DIFF
--- a/arcadia/aregion.cpp
+++ b/arcadia/aregion.cpp
@@ -1949,11 +1949,14 @@ int ARegion::CanTax(Unit *u)
 		Object *obj = (Object *) elem;
 		forlist ((&obj->units)) {
 			Unit *u2 = (Unit *) elem;
-			if (u2->guard == GUARD_GUARD && u2->IsReallyAlive())
-			    if(u2->type == U_GUARD || u2->type == U_GUARDMAGE) {
-			        if(town) return 0; //guards prevent taxing in towns
-			        if(u2->faction->ethnicity != u->faction->ethnicity) return 0; //guards prevent other ethnicities taxing
-				} else if (u2->GetAttitude(this, u) <= A_NEUTRAL) return 0;
+			if (u2->guard == GUARD_GUARD && u2->IsReallyAlive()) {
+				if (u2->type == U_GUARD || u2->type == U_GUARDMAGE) {
+					if (town) return 0; //guards prevent taxing in towns
+					if (u2->faction->ethnicity != u->faction->ethnicity) return 0; //guards prevent other ethnicities taxing
+				} else {
+					if (u2->GetAttitude(this, u) <= A_NEUTRAL) return 0;
+				}
+			}
 		}
 	}
 	
--- a/arcadia/economy.cpp
+++ b/arcadia/economy.cpp
@@ -1232,7 +1232,7 @@ int ARegion::TownHabitat()
 		if(ItemDefs[ObjectDefs[obj->type].productionAided].flags & IT_FOOD) farm++;
 		if(ObjectDefs[obj->type].productionAided == I_SILVER) inn++;
 		if(ObjectDefs[obj->type].productionAided == I_HERBS) temple++;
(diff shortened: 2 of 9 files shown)


tgerigk:
forgot to mention: only commits saying "review needed" need some more thorough review.
Commit saying "fixed" are part of the other pull request (#37).

tgerigk:
All compiler-warnings done.
Tried clangs -Weverything; errhhh ... no.

arcadia/aregion.cpp
```diff
@@ -1949,11 +1949,14 @@ int ARegion::CanTax(Unit *u)
 		Object *obj = (Object *) elem;
 		forlist ((&obj->units)) {
 			Unit *u2 = (Unit *) elem;
-			if (u2->guard == GUARD_GUARD && u2->IsReallyAlive())
-			    if(u2->type == U_GUARD || u2->type == U_GUARDMAGE) {
-			        if(town) return 0; //guards prevent taxing in towns
-			        if(u2->faction->ethnicity != u->faction->ethnicity) return 0; //guards prevent other ethnicities taxing
-				} else if (u2->GetAttitude(this, u) <= A_NEUTRAL) return 0;
```
tgerigk: Problem: Does this `else` belong to the `if` on line 1952 or the `if` on line 1953?

Comparing to standard engine aregion.cpp function canTax():
- checks for living guard (line 1952 here)
- checks for attitude worse than neutral -> return 0 (line 1956 here)
   - (false = can not tax, because unfriendly guard)
- otherwise return 1 (after for-loop)
   - (true = no guard OR friendly guard)

arcadia/aregion.cpp
```diff
@@ -1949,11 +1949,14 @@ int ARegion::CanTax(Unit *u)
...
+			if (u2->guard == GUARD_GUARD && u2->IsReallyAlive()) {
+				if (u2->type == U_GUARD || u2->type == U_GUARDMAGE) {
+					if (town) return 0; //guards prevent taxing in towns
+					if (u2->faction->ethnicity != u->faction->ethnicity) return 0; //guards prevent other ethnicities taxing
+				} else {
+					if (u2->GetAttitude(this, u) <= A_NEUTRAL) return 0;
+				}
+			}
 		}
```
tgerigk: following the given code:
1. check for living guard (line 1952)
2. check if guarding unit is an actual guard (line 1953)
3. is this region a town? -> guards + town => can not tax (return 0/false) (line1954)
4. guard are NOT the same ethnicity than wannabe taxer -> can not tax (return 0/false) (line 1955)
5. otherwise: are guards unfriendly? -> can not tax (line 1956)
   (otherwise related to 1 or to 1 AND 2?)
6. can tax (after for-loop)

Prosposed solution: 
- consider lines 1953-1955 to be additions of arcadia
- therefore match the `else` with the `if` on line 1952
- that also preserves the logic of standard engine.

tgerigk: Problem here: Does this `else` (line 1956) belong to the `if` on line 1952 or the `if`` on line 1953?

following the given code:
- check for living guard (line 1952)
- check if guarding unit is an actual guard type (line 1953)
- is this region a town? -> guards + town => can not tax (return 0/false) (line1954)
- guard are NOT the same ethnicity than wannabe taxer -> can not tax (return 0/false) (line 1955)
- otherwise: are guards unfriendly? -> can not tax (line 1956)
   (otherwise related to 1 or to 1 AND 2?)
- can tax (after for-loop)

looking at the standard engine aregion.cpp function canTax():
- checks for living guard (line 1952 here)
- checks for attitude worse than neutral -> return 0 (line 1956 here)
   (false = can not tax, because unfriendly guard)
- otherwise return 1 (after for-loop)
   (true = no guard OR friendly guard)

prosposed solution:
- consider lines 1953-1955 to be additions of arcadia
- therefore match the `else` on line 1956 with the `if` on line 1952

tgerigk:
re-doing commit messages.

PR #43, opened 2019-12-24
Take from nonexistant target crashes application

sharcashmo:
When target of TAKE command does not exist program crashes


jt-traub:
This fix is already in master. closing.

PR #44, opened 2019-12-24
Gate_Lore RANDOM LEVEL weight bug

sharcashmo:
Gate_Lore RANDOM LEVEL should use same weight limits than Gate_Lore
RANDOM. Now it uses the limits of Gate_Lore GATE, with a fallback value
of 10 when it is casted at level 2 (Gate_Lore GATE is not possible at
level 2 thus it was not contemplated).


jt-traub:
This code has been pretty significantly reworked on master branch and that code looks sane.

PR #46, opened 2019-12-24
Fix prevent_sail_through workaround

sharcashmo:
Fix a workaround on PREVENT_SAIL_THROUGH by giving the ships to a new
fleet. Now ships can only be given between compatible fleets (those in
the same shore). If target fleet has no restrictions, it will inherit
restrictions of the source fleet, if any.

Code checking for ships SailThrough check has been moved to
Object::SailThroughCheck()


jt-traub:
Already in master branch. closing.

PR #47, opened 2019-12-24
Bug with moving too early to building with BUILD command

sharcashmo:
(fixed by Artyomtrityak in New Origins)

when you're in the building, you're the owner and doing 2 orders:

destroy
build "something"

it does not work
Engine tries to destroy new building, you're trying to build and breaks

+ Building [2] : Road S.
  * Unit (69), Testets (3)

unit 61
destroy
build tower

Errors during turn:
Unit (61): BUILD: Nothing to build.

Events during turn:
Unit (61): Destroys Building [3].


jt-traub:
Already in mainline.

PR #51, opened 2019-12-25
New function ModifyTerrainFlags

sharcashmo:
(added by Artyomtrityak in New Origins)

New function that allows to modify terrain flags


jt-traub:
This is already in the main codeline. Closing.

PR #53, opened 2019-12-26
Fix using assignation instead of compare

sharcashmo:
This piece of code did an assignation instead of a comparation. This
small check is there to make lands to the very north less likely to
expand further to the north, as following check does the same with
southern lands.


jt-traub:
Changes merged into PR #189.  Closing.

PR #54, opened 2019-12-28
No trade items bugfix

sharcashmo:
When less of four trade items are enabled function
ARegion::SetupCityMarket() hangs, looking for two sell and two buy
distinct trade items


sharcashmo:
Changes is only enclosing all trade items stuff inside an _if_ clause. Review with ignore spaces better ;)

jt-traub:
Pulling this fix into PR #189 .. Also filed issue #188 to make a better fix at some point in the future.

PR #56, opened 2020-01-29
Teleport default Z-level should be determined after movement

sharcashmo:
When a unit casts Teleport without giving a z coordinate the level is
taken from unit location. As Teleport is casted after movement, z
location should be taken after movement too.


sharcashmo:
Not the correct file

PR #57, opened 2020-01-29
Bugfix teleport level

sharcashmo:
Teleport default Z-level should be determined after movement

When a unit casts Teleport without giving a z coordinate the level is
taken from unit location. As Teleport is casted after movement, z
location should be taken after movement too.


sharcashmo:
Tested in test game

artyomtrityak:
This is fixed in New Origins, we will merge it form there

PR #61, opened 2021-11-14
Implemented JSON reports

nedbrek:


nedbrek:
Looking at it now, I wish I had put in more comments... :)

artyomtrityak:
it's alright, I understand most of it

aregion.cpp
```diff
@@ -1293,25 +1386,32 @@ void ARegion::WriteExits(Areport *f, ARegionList *pRegs, int *exits_seen)
...
 			temp += AString(", ") + Population() + " peasants";
+			of.PutPairInt("population", Population());
 			if (Globals->RACES_EXIST) {
 				temp += AString(" (") + ItemDefs[race].names + ")";
+				of.PutPairStr("race", ItemDefs[race].names);
 			}
 			if (present || farsight ||
 					Globals->TRANSIT_REPORT & GameDefs::REPORT_SHOW_REGION_MONEY) {
 				temp += AString(", $") + wealth;
+				of.PutPairInt("tax", wealth);
```
artyomtrityak, line 1414: we should probably put tax 0 in else condition too

aregion.cpp
```diff
@@ -1491,13 +1604,27 @@ void ARegion::WriteReport(Areport *f, Faction *fac, int month,
...
+			bool first = true;
 			forlist (&objects) {
-				((Object *) elem)->Report(f, fac, obs, truesight, detfac,
+				Object *obj = (Object *)elem;
+				if (obj->type != O_DUMMY)
+				{
+					if (first)
+					{
+						first = false;
+						of.StartArray("objects");
```
artyomtrityak, line 1615: probably same here - let's make it non-conditional

edit.cpp
```diff
@@ -1203,7 +1203,7 @@ void Game::EditGameUnitDetails(Unit *pUnit)
 		int exit = 0;
 		Awrite(AString("Unit: ") + *(pUnit->name));		
 		Awrite(AString("Unit faction: ") +
-				*(pUnit->faction->name));
+				pUnit->faction->name());
```
artyomtrityak, line 1206: why did it change?

events.cpp
```diff
@@ -28,13 +28,14 @@
 #include <map>
 #include <algorithm>
 #include "gamedata.h"
+#include <sstream>
```
artyomtrityak, line 31: I haven't seen this std header yet, what does it give us and how important is it?

events.cpp
```diff
@@ -28,13 +28,14 @@
...
 
 FactBase::~FactBase() {
 
 }
 
 void BattleSide::AssignUnit(Unit* unit) {
-	this->factionName = unit->faction->name->Str();
+	this->factionName = unit->faction->name().Str();
```
artyomtrityak, line 38: why did it change?

events.cpp
```diff
@@ -233,3 +234,104 @@ std::string Events::Write(std::string worldName, std::string month, int year) {
...
+, inum_(inum)
+, amt_(amt)
+, reg_(reg)
+{
+}
+
+std::string ProduceEvent::str(ARegionList &regions) const
+{
+	std::ostringstream os;
+	os << uname_ << ": Produces " << ItemString(inum_, amt_) << " in " << reg_->ShortPrint(&regions) << '.';
```
artyomtrityak, line 318: we mostly using string concat, not piping, can we use most common Atlantis code style to keep it consistent?

events.cpp
```diff
@@ -233,3 +234,104 @@ std::string Events::Write(std::string worldName, std::string month, int year) {
...
+}
+
+void GiveEvent::writeJson(JsonReport &of, ARegionList &regions) const
+{
+	of.StartDict(NULL);
+	of.PutPairStr("type", reversed_ ? "Receive" : "Give");
+	of.PutPairStr("giver", gname_.c_str());
+	of.PutPairStr("receiver", tname_.c_str());
+	of.PutPairStr("abbr", ItemDefs[inum_].abr);
+	of.PutPairInt("amount", amt_);
```
artyomtrityak, line 302: we are using names with underscores, why?

faction.cpp
```diff
@@ -118,7 +118,7 @@ void Attitude::Readin(Ainfile *f, ATL_VER v)
 Faction::Faction()
 {
 	exists = 1;
-	name = 0;
+	name_ = nullptr;
```
artyomtrityak, line 121: why do we change name to name_ ?

events.cpp
valdisz: This is stringstream, basically a buffer that allows easely to construct large strings.

I'm going to use that in world events too. Will refactor them a bit.

events.cpp
valdisz: Concatenation blows up the code and is not efficient as each time strings are copied. This is standart style in C/C++

artyomtrityak: My point is that code consistency is more important than efficiency. And to be honest, efficiency at Atlantis is not an issue.

valdisz: Writing 2x code and spending 2x time for that cannot be adjusted just because it was no used before. This is super standard C code.

valdisz: There is no normal string.format type in C11 and this approach converts numbers to strings without unnecessary function calls. Also it is easy to follow what happens and don't need any boilerplate.

faction.h
```diff
@@ -203,7 +205,7 @@ class Faction : public AListElem
 	//
 	AList extraPlayers;
 	AList errors;
-	AList events;
+	std::vector<IEvent*> events_;
```
valdisz, line 208: Maybe `std::list` is better here as the number of events will vary

events.cpp
valdisz: If yoi have worked with Java or C# then there is `StringBuilder` class, which practically does the same thing as `stringstream`.
`ostringstream` just means that you can only WRITE to it and no reading can happen, i.e. exactly what we need in this case.

aregion.cpp
```diff
@@ -370,9 +370,11 @@ void ARegion::RunDecayEvent(Object *o, ARegionList *pRegs)
 	pFactions = PresentFactions();
 	forlist (pFactions) {
 		Faction *f = ((FactionPtr *) elem)->ptr;
-		f->Event(GetDecayFlavor() + *o->name + " " +
+		AString msg(GetDecayFlavor());
+		msg += *o->name + " " +
 				ObjectDefs[o->type].name + " in " +
-				ShortPrint(pRegs));
+				ShortPrint(pRegs);
+		f->LogEvent(new StrEvent(msg.Str()));
```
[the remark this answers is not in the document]
valdisz: Maybe it is better to name it `AddEvent`
I think it will better capture intent.

events.cpp
valdisz: The good point for `stringstream` is that you can put in `string` or `number` and it will be handled correctly for you. No need to call additional methods or constructs.

aregion.cpp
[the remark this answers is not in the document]
nedbrek: Part of it is my process. I try to keep the code working at all times, so I had the old function along with a new one. I can then go through and change them one by one, and everything still works. In the end, I see if I can remove the old function (which I did in the case of Faction::Event(), but not Unit::Event()).

aregion.cpp
```diff
@@ -1138,20 +1167,28 @@ int ARegion::CanMakeAdv(Faction *fac, int item)
...
 }
 
-void ARegion::WriteProducts(Areport *f, Faction *fac, int present)
+void ARegion::WriteProducts(JsonReport &of, Faction *fac, int present)
 {
+	Areport *f = of.PlainFile();
 	AString temp = "Products: ";
 	int has = 0;
+	bool has_js = false;
+	int ent_amt = 0;
```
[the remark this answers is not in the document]
nedbrek: I've had some bugs in class JsonReport with empty arrays and dictionaries. I should fix that, but this was easy for now :)

aregion.cpp
```diff
@@ -1160,6 +1197,7 @@ void ARegion::WriteProducts(Areport *f, Faction *fac, int present)
 							GameDefs::REPORT_SHOW_ENTERTAINMENT) || present) {
 						f->PutStr(AString("Entertainment available: $") +
 								p->amount + ".");
+						ent_amt = p->amount;
```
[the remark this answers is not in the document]
nedbrek: That sounds good.

events.cpp
nedbrek: For JSON, I print the faction name and number separately (currently, just in the faction status, but it could be done everywehre). The old name field had both, now name_ member just as the raw name, and name() method returns the name and number.

events.cpp
nedbrek: It's a pattern I've found very helpful. Class data has underscore, so you can easily tell if a class method is altering local variables or updating class state.

faction.h
nedbrek: On a modern processor, vector will outperform list in almost every case.

aregion.cpp
```diff
@@ -1138,20 +1167,28 @@ int ARegion::CanMakeAdv(Faction *fac, int item)
...
 	int has = 0;
+	bool has_js = false;
+	int ent_amt = 0;
 	forlist((&products)) {
 		Production *p = ((Production *) elem);
 		if (ItemDefs[p->itemtype].type & IT_ADVANCED) {
 			if (CanMakeAdv(fac, p->itemtype) || (fac->IsNPC())) {
 				if (has) {
-					temp += AString(", ") + p->WriteReport();
+					temp += AString(", ");
```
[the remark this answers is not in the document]
nedbrek: This is part of the refactoring I missed when backing out some of these changes. We can back this one out too.

aregion.cpp
```diff
@@ -1170,16 +1208,26 @@ void ARegion::WriteProducts(Areport *f, Faction *fac, int present)
 						GameDefs::REPORT_SHOW_RESOURCES))
 					continue;
 				if (has) {
-					temp += AString(", ") + p->WriteReport();
+					temp += AString(", ");
```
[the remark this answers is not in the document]
nedbrek: It's down on line 1220 like above.

aregion.cpp
```diff
@@ -1242,48 +1299,84 @@ void ARegion::WriteMarkets(Areport *f, Faction *fac, int present)
...
+void ARegion::WriteEconomy(JsonReport &of, Faction *fac, int present)
 {
+	Areport *f = of.PlainFile();
 	f->AddTab();
 
+	double wages = 0;
+	int max_wages = 0;
 	if ((Globals->TRANSIT_REPORT & GameDefs::REPORT_SHOW_WAGES) || present) {
-		f->PutStr(AString("Wages: ") + WagesForReport() + ".");
+		Production *p = products.GetProd(I_SILVER, -1);
```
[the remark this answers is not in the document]
nedbrek: Yeah, I need to capture some of the internal calculations and didn't want to mess around adding a bunch of reference variables - so I just inlined the function :)

events.cpp
nedbrek: Yeah, it's part of modernizing the code. In Miskatonic, I've started bringing in some more modern concepts like templates and lambda functions. It makes the code a lot easier to read.

nedbrek:
I did an amend with force push, so those deltas in WriteReport() don't even show now.

aregion.cpp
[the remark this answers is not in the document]
nedbrek: Fixed

aregion.cpp
```diff
@@ -1242,48 +1299,84 @@ void ARegion::WriteMarkets(Areport *f, Faction *fac, int present)
...
 
-void ARegion::WriteExits(Areport *f, ARegionList *pRegs, int *exits_seen)
+void ARegion::WriteExits(JsonReport &of, ARegionList *pRegs, int *exits_seen)
 {
+	Areport *f = of.PlainFile();
 	f->PutStr("Exits:");
 	f->AddTab();
-	int y = 0;
+
+	bool seen_exits = false;
```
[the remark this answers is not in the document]
nedbrek: Fixed

aregion.cpp
nedbrek: Fixed

runorders.cpp
```diff
@@ -3027,14 +3037,14 @@ int Game::DoGiveOrder(ARegion *r, Unit *u, GiveOrder *o)
 	}
 
 	if (o->type == O_TAKE) {
+		// TODO JSON
```
artyomtrityak, line 3040: TODO?

unit.cpp
```diff
@@ -535,17 +628,97 @@ void Unit::WriteReport(Areport *f, int obs, int truesight, int detfac,
...
+				continue;
+
+			if (SkillDefs[sk].flags & SkillType::FOUNDATION)
+			{
+				// This is only used with mages to display studiable magic skills
+				if (type != U_MAGE)
+					continue;
+
+				// Compare current skill level to maximum level
+				// TODO find LCM man
```
artyomtrityak, line 713: is there anything here needed to be finished?

runorders.cpp
nedbrek: Only if we want a specialized TAKE Json report. It will show up now as plain text.

unit.cpp
nedbrek: Only if you want to show "Can Study" for non-mages.

jt-traub:
This has been superceded by the JSON work which has already been added to the master branch.

Issue #71, opened 2021-11-19
Merge WAR and TRADE into Empire. Empire | Magic 5 FP


valdisz:

jt-traub:
This was the martial code. closing issue.

PR #79, opened 2021-11-21
Martial faction points

valdisz:
This PR merges War and Trade points into Martial points. Activity tracking is made more flexible in the code.
Faction points are defined during runtime and are using vector.

#71

### DIFF
--- a/faction.cpp
+++ b/faction.cpp
@@ -25,6 +25,7 @@
 
 #include "gamedata.h"
 #include "game.h"
+#include <algorithm>
 
 char const *as[] = {
 	"Hostile",
@@ -36,13 +37,13 @@ char const *as[] = {
 
 char const **AttitudeStrs = as;
 
-char const *fs[] = {
-	"War",
-	"Trade",
-	"Magic"
-};
+const std::string F_WAR = "War";
+const std::string F_TRADE = "Trade";
+const std::string F_MAGIC = "Magic";
+const std::string F_MARTIAL = "Martial";
 
-char const **FactionStrs = fs;
+std::vector<std::string> ft { };
+std::vector<std::string> *FactionTypes = &ft;
 
 // LLS - fix up the template strings
 char const *tp[] = {
@@ -119,9 +120,11 @@ Faction::Faction()
 {
 	exists = 1;
 	name = 0;
-	for (int i=0; i<NFACTYPES; i++) {
-		type[i] = 1;
+	
+	for (auto &ft : *FactionTypes) {
+		type[ft] = 1;
 	}
+	
 	lastchange = -6;
 	address = 0;
 	password = 0;
@@ -141,9 +144,11 @@ Faction::Faction(int n)
 {
 	exists = 1;
 	num = n;
-	for (int i=0; i<NFACTYPES; i++) {
-		type[i] = 1;
+	
+	for (auto &ft : *FactionTypes) {
+		type[ft] = 1;
 	}
+
 	lastchange = -6;
 	name = new AString;
 	*name = AString("Faction (") + AString(num) + AString(")");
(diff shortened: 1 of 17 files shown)


standard/rules.cpp
```diff
@@ -235,8 +240,11 @@ static GameDefs g = {
 	0, // STRICT_GUARD
 	1, // OCEAN_GUARD
 	BattleLogLevel::NORMAL,	// BATTLE_LOG_LEVEL
-	1,	// WORLD_EVENTS
+	0,	// EXTENDED_FORT_DEFENCE
```
valdisz, line 243: many new options were missing from standard

faction.cpp
```diff
@@ -36,13 +37,13 @@ char const *as[] = {
 
 char const **AttitudeStrs = as;
 
-char const *fs[] = {
-	"War",
-	"Trade",
-	"Magic"
-};
+const std::string F_WAR = "War";
```
artyomtrityak, line 40: why do we changing this to std string?

faction.cpp
```diff
@@ -927,3 +951,37 @@ void Faction::DiscoverItem(int item, int force, int full)
...
+
+	return count;
+}
+
+void Faction::RecordActivity(ARegion *region, FactionActivity type) {
+	this->activity[region].insert(type);
+}
+
+bool Faction::IsActivityRecorded(ARegion *region, FactionActivity type) {
+	auto regionActivity = this->activity[region];
```
artyomtrityak, line 985: was it pushed "as it is" or changes to auto?

faction.h
```diff
@@ -52,12 +56,10 @@ enum {
 	NATTITUDES
 };
 
-enum {
-	F_WAR,
-	F_TRADE,
-	F_MAGIC,
-	NFACTYPES
-};
+extern const std::string F_WAR;
```
artyomtrityak, line 59: if we change FWAR to enum, can we keep the same AString?

fracas/rules.cpp
```diff
@@ -58,6 +58,11 @@ static int ag[] = { 0, 1, 2, 4, 6, 10 };
 int *allowedTacticians = ag;
 int allowedTacticiansSize = sizeof(ag) / sizeof(ag[0]);
 
+// allowed Martial activity
+static int ma[] = { 0, 10, 25, 40, 60, 90 };
```
artyomtrityak, line 62: why do we need this here if it is off?

artyomtrityak (commented):
looks good, did you run tests in a local game?

fracas/rules.cpp
valdisz: Code must compile for all rulesets, right? :)

faction.h
valdisz: I would be more afraid of using `AString` instead of `std::string`.

The enum will make code more complicated because it will need those constants anyway (to be used in parsing) and it will need mapping.

faction.cpp
valdisz: So that there is no unexpected behavior in using strings with STD containers. `std::string` is better in all aspects and there are no problems to convert it into `AString`.

`std::string` is available since 1998. It existed already when Atlantis code was written. `AString` should not exist then, and it is wrong to use it now.

faction.cpp
valdisz: `IsActivityRecorded` is a new function.

valdisz: `auto` does not mean that variable will have a dynamic type or something. It just tells that compiler should infer type during compilation. It is the same as `var` in C# or Java.

No need to write long type names if the compiler knows them.

arriveby:
And look, 7 mages for Fracas ;)

valdisz:
> looks good, did you run tests in a local game?

Yes, was working fine. Need to admit that this was not quite big test game.

artyomtrityak (approved):

PR #80, opened 2021-11-21
Non-instant building destruction

valdisz:
There are 3 options for how buildings can be destroyed:
1. Instantly (standard)
2. Per figure - one figure will contribute to removing one structural point
3. Per skill - each figure will contribute to remove as many structural points as it knows `BUIL` skill

When `PER_FIGURE` or `PER_SKILL` is used, max % of removed structural points (from the complete building) and min amount of removed points, can be specified. Such way small structures can be removed in one turn, but larger structures would need several turns even if w whole several thousand army is performing destrucion.

This solves #69

### DIFF
--- a/fracas/rules.cpp
+++ b/fracas/rules.cpp
@@ -248,6 +248,9 @@ static GameDefs g = {
 	1,	// WORLD_EVENTS
 	0,	//FACTION_STATISTICS
 	FactionActivityRules::DEFAULT,	// FACTION_ACTIVITY
+	DestroyBehavior::INSTANT,	// DESTROY_BEHAVIOR
+	200,	// MIN_DESTROY_POINTS,
+	34,	// MAX_DESTROY_PERCENT
 };
 
 GameDefs *Globals = &g;
--- a/gamedefs.h
+++ b/gamedefs.h
@@ -80,6 +80,12 @@ enum FactionActivityRules {
 	MARTIAL_MERGED = 2	// Common point pool for War and Trade activity, but counted per region - do what you want in a region.
 };
 
+enum DestroyBehavior {
+	INSTANT    = 0,	// default behavior,any unit can instantly destroy any building
+	PER_FIGURE = 1,	// one man will destroy one building strucure point, it is assumed that destroy power is 1
+	PER_SKILL  = 2	// use building skill as a basis how much can be destroyed. The formula is: destroy power = max(1, BUILDING skill level)
+};
+
 class GameDefs {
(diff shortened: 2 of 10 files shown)


valdisz:
@artyomtrityak

fracas/rules.cpp
```diff
@@ -248,6 +248,9 @@ static GameDefs g = {
 	1,	// WORLD_EVENTS
 	0,	//FACTION_STATISTICS
 	FactionActivityRules::DEFAULT,	// FACTION_ACTIVITY
+	DestroyBehavior::INSTANT,	// DESTROY_BEHAVIOR
+	200,	// MIN_DESTROY_POINTS,
+	34,	// MAX_DESTROY_PERCENT
```
artyomtrityak, line 253: Why do we want to limit it to 34 instead of 100?

gamedefs.h
```diff
@@ -771,6 +777,18 @@ class GameDefs {
...
+	// how buildings are destroyed, read more in DestroyBehavior enum
+	DestroyBehavior DESTROY_BEHAVIOR;
+
+	// set folowing value when DESTROY_BEHAVIOR is not INSTANT
+	// how much structure points can be destroyed per turn
+	// MAX_DESTROY_PERCENT accepts values from [1..100]
+	// MIN_DESTROY_POINTS accepts [0..infinity]
+	// formula:
+	//     turn destroyable points = max(MIN_DESTROY_POINTS, Max Structure Points * MAX_DESTROY_PERCENT / 100)
+	int MIN_DESTROY_POINTS;
```
artyomtrityak, line 790: can you please explain about min destroy points and why they needed?

artyomtrityak (approved):

fracas/rules.cpp
valdisz: In that particular ruleset it won't be used because instant destruction is turned on. I just copied the same values around.

gamedefs.h
valdisz: If destruction is not instant then it helps to control how fast buildings can be destroyed. 

You can make so that small buildings can be destroyed in a turn, but huge one's will need several turns no matter what.

Current example with 34% and min 200 point destroyble points will make citadel required 3 turns to fully destroyed,  castle 2 turns, and the rest of the structures will need 1 turn. if sufficient man count of course.

valdisz: Setting % to 100 will make all buildings destroyeable in a turn

PR #86, opened 2021-12-19
Half riding bonus

artyomtrityak:


army.cpp
```diff
@@ -354,6 +355,16 @@ Soldier::Soldier(Unit * u,Object * o,int regtype,int r,int ass)
 				break;
 			}
 		}
+
+		// Modify riding bonus for half bonus
+		if (Globals->HALF_RIDI_BONUS) {
+			printf("\n\n RIDING before: %d\n", ridingBonus);
```
artyomtrityak, line 361: will delete after testing

army.cpp
```diff
@@ -354,6 +355,16 @@ Soldier::Soldier(Unit * u,Object * o,int regtype,int r,int ass)
...
+
+			ridingBonus = ceil(ridingBonus / 2);
```
valdisz, line 363: `int` divided by `int` will be `int` and `ceil` will not work in such a way, you would need to cast at least one of the sides to one of the floating-point types.

Better do it like this, then no need for `ceil` and floating-point operations: `(ridingBonus + 1) / 2`

It will produce such results:
```
RIDI Level    Bonus
0             0
1             1
2             1
3             2
4             2
5             3
```

valdisz (changes requested):

valdisz (dismissed):

PR #97, opened 2022-01-29
New map generator

valdisz:


aregion.cpp
```diff
@@ -2743,6 +2748,718 @@ int mapBiome(int biome) {
...
+	for (auto wb : list) {
+		if (isNearWaterBody(reg, wb)) {
+			return true;
+		}
+	}
+
+	return false;
+}
+
+void makeRivers(Map* map, ARegionArray* arr, std::vector<WaterBody*>& waterBodies, std::unordered_map<ARegion*, int>& rivers,
```
artyomtrityak, line 2835: should it go to mapgen instead of region file?

neworigins/.gitignore
```diff
@@ -0,0 +1,4 @@
+node_modules/
```
artyomtrityak, line 1: we should delete all js/ts/json files

aregion.cpp
valdisz: then there will be a circular dependency

neworigins/.gitignore
valdisz: when this is done
I committed map rendering to allow others see the map w/o client while development happens

artyomtrityak (approved):

PR #98, opened 2022-01-31
Update resources

artyomtrityak:


gamedata.cpp
```diff
@@ -4261,7 +4261,7 @@ static TerrainType td[] = {
 	 TerrainType::RIDINGMOUNTS | TerrainType::FLYINGMOUNTS | TerrainType::SHOW_RULES,
 	 100,11,10,2,
 	 {{I_FUR,100,10},{I_HERBS,100,10},{I_MUSHROOM,35,5},{I_IRON,30,5},
-	  {-1,0,0},{-1,0,0},{-1,0,0}},
+	  {I_WHORSE,35,6},{-1,0,0},{-1,0,0}},
```
valdisz, line 4264: Do we have war horses?

artyomtrityak: ah yeah it should be WING :)

PR #103, opened 2022-02-06
PR adds SKILLOUT_PROGRESSIVE and changes food to 1/3/5 1/2/3

artyomtrityak:


items.h
```diff
@@ -96,7 +96,8 @@ class ItemType
 			// This item cannot be transported.
 			NOTRANSPORT = 0x20,
 			// Produced monsters
-			MANPRODUCE = 0x40
+			MANPRODUCE = 0x40,
+			SKILLOUT_PROGRESSIVE = 0x50
```
valdisz, line 100: We have similar feature in other skills and it is named differently. From the name "progressive" is not clear what it does.

valdisz (changes requested):

monthorders.cpp
```diff
@@ -1178,6 +1180,15 @@ void Game::RunUnitProduce(ARegion * r,Unit * u)
...
 		output *= level;
+	if (ItemDefs[o->item].flags & ItemType::SKILLOUT_PROGRESSIVE) {
+		if (level == 3) {
+			output *= 2;
+		}
+		if (level == 5) {
+			output *= 3;
+		}
+	}
+
```
valdisz, line 1191: What you wanted to achieve here? To have skill level rounded up produced amount just do: (level + 1) / 2

Then results will be following:
L1 = 1 item
L2 = 1 item
L3 = 2 items
L4 = 2 items
L5 = 3 items

`output *= (level + 1) / 2`

skillshows.cpp
```diff
@@ -1631,6 +1631,10 @@ AString *ShowSkill::Report(Faction *f)
 				if (ItemDefs[i].flags & ItemType::SKILLOUT) {
 					temp1 += " equal to their skill level";
 				}
+				if (ItemDefs[i].flags & ItemType::SKILLOUT_PROGRESSIVE) {
+					temp1 += " based on their skill level. Level 1 produces 1 item, ";
+					temp1 += " level 3 produces 2 items, level 5 produces 3 items";
```
valdisz, line 1636: Then here is easy to make generic examples w/o hardcoding.

valdisz (changes requested):

items.h
valdisz: This is an output multiplier. Maybe use flag name that reflects such behaviour?

artyomtrityak: sure, I will change name

monthorders.cpp
artyomtrityak: sounds good

valdisz (approved):

Issue #102, opened 2022-02-03
Wrong Martial faction point description in the rules


valdisz:

jt-traub:
Not sure why the associated PR with this was closed, but there are some changes in that PR which are still valid.  Adding this issue to my board as well so I don't lose track of it.

jt-traub:
These changes along with the ones from PR #108  (which mostly overlapped) have been applied and this issue is now closed.
See issue #106  and PR #192

PR #105, opened 2022-02-07
Update genrules.cpp

Elfinhope:
Updated genrules.cpp to reflect change from War/Trade/Magic faction types to Martial/Magic.  Included some minor grammatical changes as well.  Changes also made to Combat section rules for movement after battle.  This addresses issues #102 and #106.

### DIFF
--- a/genrules.cpp
+++ b/genrules.cpp
@@ -171,21 +171,21 @@ void Game::WriteFactionTypeDescription(std::ostringstream& buffer, Faction &fac)
 				buffer << "perform tax or trade in " << nma << " " << plural(nma, "region", "regions");
 			}
 			
-			if (Globals->FACTION_ACTIVITY == FactionActivityRules::MARTIAL_MERGED) { 
+			if (Globals->FACTION_ACTIVITY == FactionActivityRules::MARTIAL_MERGED) {
 				buffer << "perform tax and trade in " << nma << " " << plural(nma, "region", "regions");
 			}
 		}
 
 		count++;
 	}
-
+/*
 	if (Globals->APPRENTICES_EXIST && na > 0) {
 		buffer << ", as well have " << na << " " << Globals->APPRENTICE_NAME;
 		if (na > 1) {
 			buffer << "s";
 		}
 	}
-
+*/
 	if (Globals->TRANSPORT & GameDefs::ALLOW_TRANSPORT && nq > 0) {
 		buffer << ", and " << nq << " quartermaster";
 		if (nq > 1) {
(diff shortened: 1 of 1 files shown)


genrules.cpp
```diff
@@ -171,21 +171,21 @@ void Game::WriteFactionTypeDescription(std::ostringstream& buffer, Faction &fac)
...
-			if (Globals->FACTION_ACTIVITY == FactionActivityRules::MARTIAL_MERGED) { 
+			if (Globals->FACTION_ACTIVITY == FactionActivityRules::MARTIAL_MERGED) {
 				buffer << "perform tax and trade in " << nma << " " << plural(nma, "region", "regions");
 			}
 		}
 
 		count++;
 	}
-
+/*
```
artyomtrityak, line 181: why did you commented this part?

genrules.cpp
```diff
@@ -3064,14 +3082,20 @@ int Game::GenRules(const AString &rules, const AString &css,
...
+			"directly, but when completed are placed in Fleets. "
+			"Fleets may contain one or more ships, and "
 			"may be entered like other buildings.";
 		f.Paragraph(temp);
 		temp = "";
 		if (Globals->FACTION_LIMIT_TYPE == GameDefs::FACLIM_FACTION_TYPES) {
-			temp += "Only factions with at least one faction point spent on "
-				"trade can issue ";
+			if (Globals->BUILD_NO_TRADE) {
+				temp = +"Any faction ";
```
gelzis, line 3092: I guess this was intended as `temp += "Any faction ";`

genrules.cpp
Elfinhope: I commented that out a couple of days ago...When it generated the rules it was printing "as well as 5 apprentices" twice. Once with the information about mages and a second time at the end of the description block.  Commenting out this code got rid of the second time.  I left it in but commented out as I was not sure what effect that change would have under different rules configurations.  It works properly with current configuration, but haven't had time to make sure that the output is correct with Martial (not Martial Merged) and original War/Trade/Magic rules.

genrules.cpp
Elfinhope: Yes not sure how that compiled that way. I'm pretty sure I checked the output and it was correct... I'll double check it tonight when I get off work.

Elfinhope: This has been addressed and patch pushed.

artyomtrityak:
@Elfinhope is it ready to review / merge?

valdisz (approved):

Issue #109, opened 2022-02-11
Clarify resources consumed with CONSUME order in rules


Elfinhope:
Rules are vague as to exactly what is consumed and in what order when the CONSUME order is issued to a unit.

rmstdope:
@Elfinhope, please review my pull request mentioned above. Does this fully resolve this issue or do you believe we need to clarify things further?

PR #111, opened 2022-02-21
Added clarification on how maintenance fee is paid

rmstdope:
Added a priority list of how the maintenance fee is paid for a unit. Addresses issue #109.

### DIFF
--- a/genrules.cpp
+++ b/genrules.cpp
@@ -2473,7 +2473,7 @@ int Game::GenRules(const AString &rules, const AString &css,
 			if (ItemDefs[i].flags & ItemType::DISABLED) continue;
 			if (!(ItemDefs[i].type & IT_FOOD)) continue;
 			if (last != -1) {
-				if (j > 0) temp += ", ";
+				if (j > 1) temp += ", ";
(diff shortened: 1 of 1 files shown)


Elfinhope:
Yes it looks like this resolves the issue.  Thanks.  I haven't had time to research the issue and get a fix put together.

genrules.cpp
```diff
@@ -2501,8 +2501,51 @@ int Game::GenRules(const AString &rules, const AString &css,
 		temp += "these items are worth more when sold in towns, so selling "
 			"them and using the money is more economical than using them for "
 			"maintenance.";
-	};
-	f.Paragraph(temp);
+		f.Paragraph(temp);
+
+		temp = "In summary, the order of priority for how the maintenance fee is paid "
```
valdisz, line 2506: I think `Here is how the maintenance fee is paid:` would be better. But still, there is the passive voice...

valdisz (changes requested):

jt-traub:
Pulled into PR#189.  Closing

PR #112, opened 2022-02-21
Updated description for FORGET to reflect the defined ruleset

rmstdope:
Ensured we check SKILL_LIMIT_NONLEADERS before stating that normal units can know only one skill.
Added another example when FORGET may come in handy.

### DIFF
--- a/genrules.cpp
+++ b/genrules.cpp
@@ -5049,8 +5049,26 @@ int Game::GenRules(const AString &rules, const AString &css,
 	f.ClassTagText("div", "rule", "");
 	f.LinkRef("forget");
 	f.TagText("h4", "FORGET [skill]");
-	temp = "Forget the given skill. This order is useful for normal units "
-		"who wish to learn a new skill, but already know a different skill.";
+	temp = "Forget the given skill. This order is useful for ";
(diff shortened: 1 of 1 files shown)


genrules.cpp
```diff
@@ -5049,8 +5049,26 @@ int Game::GenRules(const AString &rules, const AString &css,
 	f.ClassTagText("div", "rule", "");
 	f.LinkRef("forget");
 	f.TagText("h4", "FORGET [skill]");
-	temp = "Forget the given skill. This order is useful for normal units "
-		"who wish to learn a new skill, but already know a different skill.";
+	temp = "Forget the given skill. This order is useful for ";
+	if (Globals->SKILL_LIMIT_NONLEADERS) {
+		temp += "normal units who wish to learn a new skill, but already "
+			"know a different skill and for ";
```
artyomtrityak, line 5055: we probably can rephrase it somehow / split it into multiple sentences? It seems long hard to read sentence

rmstdope: Agree. I have split it into three sentences, which makes it a bit easier to read.

rmstdope:
This one is ready for review again after latest commit.

artyomtrityak (approved):

PR #114, opened 2022-03-01
display effect damage if only total dealt damage is more than 0

valdisz:


artyomtrityak:
We still want to display when fireball does 0 damage so we should exclude magic from this.
Though this is probably harder issue that we need to avoid spam but we need to have some details

valdisz:
there will be info about all fireball attempts in the battle stats anyway

artyomtrityak (approved):

PR #121, opened 2022-09-23
Show unit consumption in events

valdisz:
What we will see:
1) how much unit spent on study
2) how much unit spent on maintenance
3) how much unit consumed food

```
Events during turn:
Unit (157): Produces 14 livestock [LIVE] in swamp (10,12) in
  Ektromakos Marshes.
Unit (217): Earns 13 silver working in plain (17,15) in Amalirr
  Valley.
Unit (157): Consumes 28 livestock [LIVE] for maintenance.
Unit (157): Consumes 160 silver [SILV] for maintenance.
Unit (217): Consumes 50 silver [SILV] for maintenance.
```


jt-traub:
Incorporated in #187

PR #127, opened 2022-12-27
Update to use C++20 (not C++11)

jt-traub:
Also, removed all warnings to make compile clean.

Primarily this consisted of fixing type comparisons between int and size_t types. There were a couple of cases where missing braces led to warnings for ifs, including at least one that I suspect was a latent bug should the options have ever been enabled.

Lastly there was a change in the std istream that required a simple change in Astring.cpp to avoid an issue when moving from C++17 to C++20.

CLang on macOS was even pickier and complained about some things that g++ didn't.  Fixed those warnings too.
This included an error where the std::shuffle_random() was removed (as of C++17) and replaced with a more deterministic
and higher quality rng alternative but explicit rather than implicit.

### DIFF
--- a/aregion.cpp
+++ b/aregion.cpp
@@ -3007,7 +3007,10 @@ void makeRivers(Map* map, ARegionArray* arr, std::vector<WaterBody*>& waterBodie
 					continue;
 				}
 
-				graph.setInclusion([ source, target, &rivers, &waterBodies ](ARegion* current, ARegion* next) {
+				// Commented out waterBodys to match that it's unused currently due
+				// to being commented out.
+				// TODO: this probably should be a game config for that check
+				graph.setInclusion([ source, target, &rivers /*, &waterBodies */ ](ARegion* current, ARegion* next) {
 					if (source->includes(next)) {
 						// river can't go through source water body
 						return false;
@@ -3254,7 +3257,7 @@ std::vector<graphs::Location2D> getPoints(const int w, const int h,
 	
 	graphs::Location2D loc;
 	do {
-		loc = { x: getrandom(w), y: getrandom(h) };
+		loc = { .x = getrandom(w), .y = getrandom(h) };
 	}
 	while (!onIsIncluded(loc));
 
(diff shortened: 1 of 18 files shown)


i_rand.cpp
```diff
@@ -44,99 +44,146 @@ rand.c: By Bob Jenkins.  My random number generator, ISAAC.
...
-	*(r++) = b = ind(mm,y>>RANDSIZL) + x; \
-}
+#define ind(mm, x) (*(ub4 *)((ub1 *)(mm) + ((x) & ((RANDSIZ - 1) << 2))))
+#define rngstep(mix, a, b, mm, m, m2, r, x)      \
+	{                                            \
+		x = *m;                                  \
+		a = (a ^ (mix)) + *(m2++);               \
+		*(m++) = y = ind(mm, x) + a + b;         \
+		*(r++) = b = ind(mm, y >> RANDSIZL) + x; \
+	}
```
jt-traub, line 54: This is just VSCode being opinionated about indentation.. no function change here.

valdisz:
What good things come with `C++20`?

valdisz (dismissed):

valdisz:
@jt-traub the Mac OS build is failing :(

jt-traub:
> What good things come with `C++20`?

C++ has evolved a fair bit over the years and it seemed silly to lock the code out of new features around generics, etc which the newer libraries support (in the hopes that we can over-time modernize the code base and get rid of silliness like having a string class instead of using the stdlib ones).. Same for lists (collections), file io, etc.. For now, it does nothing except open that door.. we still need to make the effort of updating the code.

jt-traub:
> @jt-traub the Mac OS build is failing :(

Yeah.. I see that.. need to get ahold of a macOs machine and dig into it.. I wasn't expecting that so don't know why it might be until I look deeper.

aregion.cpp
```diff
@@ -3734,13 +3737,13 @@ void economy(ARegionArray* arr, const int w, const int h) {
...
-				addSettlement: false,
-				settlementName: std::string(),
-				settlementSize: 0
+				.terrain = terrain,
+				.habitat = terrain->pop + 1,
+				.prodWeight = 1,
+				.addLair = addLair,
+				.addSettlement = false,
+				.settlementName = std::string(),
+				.settlementSize = 0
```
jt-traub, line 3746: all of these are due to CLang complaining about new-style object initialization (what I change it to) versus and older gnu-style initialization (what was there).

astring.cpp
```diff
@@ -118,22 +118,22 @@ AString & AString::operator=(const char *c)
 	return *this;
 }
 
-int AString::operator==(char *s)
+int AString::operator==(char *s) const
```
jt-traub, line 121: clang didn't like comparing these and complained of ambiguity depending on which one was chose as the rhs/lhs.  This is the accepted way of doing this nowadays.

valdisz:
@jt-traub is this done?

jt-traub:
@valdisz  yes.  I just don't have the permission to complete the merge, so was waiting for you or @artyomtrityak  to do so.

artyomtrityak:
I will run local tests

PR #129, opened 2023-01-15
effect spells should not have damage

valdisz:


artyomtrityak:
Is this PR still valid?

valdisz:
Yes, it is still valid. Fear (and others) does not deal any damage. Also empty lines... are empty now :)

jt-traub:
Incorporated in #187

PR #133, opened 2023-09-12
Orders checker checks that password is valid.

sheckman16:
At present, the orders checker does not check passwords. So players discover password errors not when they submit, but when they get their reports.  (I think) this change makes the orders checker check passwords.


jt-traub:
Closing as this was fundamentally identical to #134 and the change has been captured in my branch to clean up old PRs.

jt-traub:
Changes are in PR #186

PR #136, opened 2023-10-04
Allow @endturn on input orders

jt-traub:
This allows the following structure
@turn
    <some orders>
@endturn

to not end up screwing up and thinking you failed to end your orders.   The checker and ALH already thought this was
fine.  This just brings the game-processing in-line.

### DIFF
--- a/parseorders.cpp
+++ b/parseorders.cpp
@@ -2042,6 +2042,8 @@ AString *Game::ProcessTurnOrder(Unit *unit, Aorders *f, OrdersCheck *pCheck,
 			order = new AString("#end");
 		}
 		AString	saveorder = *order;
+		// In order to allow @endturn to work the same as endturn we need to check for and eat the possible @
+		int getatsign = order->getat();
 		token = order->gettoken();
 
 		if (token) {


artyomtrityak (approved):
Thank you for contribution

parseorders.cpp
```diff
@@ -2042,6 +2042,8 @@ AString *Game::ProcessTurnOrder(Unit *unit, Aorders *f, OrdersCheck *pCheck,
 			order = new AString("#end");
 		}
 		AString	saveorder = *order;
+		// In order to allow @endturn to work the same as endturn we need to check for and eat the possible @
+		int getatsign = order->getat();
```
artyomtrityak, line 2046: Does it also eats turn `@` ?
It changes it for every order, is it intended to be?

artyomtrityak:
I am not sure it is correct solution.
We should change ```code = Parse1Order(token);``` in `void Game::ParseOrders` so it will parse correctly `@endturn` orders

artyomtrityak (changes requested):

parseorders.cpp
jt-traub: It doesn't change it for every order.    The order that gets copied into the turn save, is a copy (saveorder). This modifies the local order, but that isn't used except for checking for the end conditions for the turrn/endturn pair.   Those were in fact broken if any of those ending conditions were preceeded by an @ leading to the buggy behavior.   The code as it is now will just eat the @ before parsing the order and testing it.   I don't think parsing the @ in Parse1Order will work given that the outer loop for this (the main parse loop) also relies on pulling off the @ and parsing the order, so you would need to return multiple values from Parse1Order which today just returns the int value of the current order it sees under it's cursor.

I tested this locally to make sure it didn't break anything by locally running turns with and without the @ in front of both the external and internal commands and they generated identical output now in all cases which is what we want.

In the order template output, it *will* be converted from @turn/@endturn to the @turn/endturn (without the @) which is also right.  This is basically following the principle of 'be liberal in the input you accept, and be strict in the output you generate'.

jt-traub: See https://github.com/Atlantis-PBEM/Atlantis/blob/neworigins-v6/astring.cpp#L102-L119 specifically lines 107 and 117

artyomtrityak (approved):

PR #138, opened 2023-10-05
Clean up compiler warnings for all platforms

jt-traub:
We have a number of compiler warnings in the current code.   Let's
get rid of them all.

Add -Werror to prevent new compile warnings from creeping in


valdisz (approved):

valdisz:
LGTM
@artyomtrityak it is just a syntax change on 80% If no objections I will merge it in a week

PR #139, opened 2023-10-09
Include boost-ext/ut micro unit test library

jt-traub:
This adds the [Boost μt](https://github.com/boost-ext/ut#ut--%CE%BCt) unit testing framework to the codebase.

This is a header-only freely redistributable framework which encapsulates most different unit testing styles (suites, rspec, gherkin).  For the tests here, I am using the most simple 'suites' style, but we could use other styles as well.

The overall engine however, is not very amenable to unit testing in it's current state.   I have provided some examples of testing various functions within the faction and unit classes, in order to provide some guidance on how tests might be written, but writing thorough tests will be hard without refactoring a lot of the current engine framework.    That is something I believe should be undertaken over time however.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -12,7 +12,7 @@ on:
 
 jobs:
   compile:
-    name: Complie
+    name: Compile
     runs-on: ${{ format('{0}-latest', matrix.os) }}
     strategy:
       matrix:
@@ -24,9 +24,9 @@ jobs:
            - os: windows
              ext: .exe
     steps:
-      - uses: actions/checkout@v2
+      - uses: actions/checkout@v4
       - run: make all-clean && make all
-      - uses: actions/upload-artifact@v2
+      - uses: actions/upload-artifact@v3
         with:
           name: ${{ matrix.os }}
           path: |
@@ -42,9 +42,18 @@ jobs:
     runs-on: ubuntu-latest
     needs: compile
     steps:
-      - uses: actions/checkout@v2
-      - uses: actions/download-artifact@v2
+      - uses: actions/checkout@v4
+      - uses: actions/download-artifact@v3
         with:
           name: ubuntu
       - run: ./run-snapshots.sh
         working-directory: snapshot-tests
+
+  unittest:
+    name: Unit Tests
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - run: make unittest
+      - run: ./unittest
+        working-directory: unittest
--- a/Makefile
+++ b/Makefile
@@ -11,7 +11,7 @@ GAME ?= standard
 
 CPLUS = g++
 CC = gcc
-CFLAGS = -g -I. -I.. -Wall -Werror -std=c++11
+CFLAGS = -g -I. -I.. -Wall -Werror -std=c++20
 
 RULESET_OBJECTS = extra.o map.o monsters.o rules.o world.o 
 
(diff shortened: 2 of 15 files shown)


.github/workflows/build.yml
```diff
@@ -12,7 +12,7 @@ on:
 
 jobs:
   compile:
-    name: Complie
+    name: Compile
```
jt-traub, line 15: Just a dumb typo I spotted.

.github/workflows/build.yml
```diff
@@ -24,9 +24,9 @@ jobs:
            - os: windows
              ext: .exe
     steps:
-      - uses: actions/checkout@v2
+      - uses: actions/checkout@v4
```
jt-traub, line 27: The github actions for checkout and upload/download artifacts had a version bump.  I just brought them up to current.

.github/workflows/build.yml
```diff
@@ -42,9 +42,18 @@ jobs:
...
-      - uses: actions/checkout@v2
-      - uses: actions/download-artifact@v2
+      - uses: actions/checkout@v4
+      - uses: actions/download-artifact@v3
         with:
           name: ubuntu
       - run: ./run-tests.sh
         working-directory: tests
+
+  unittest:
```
jt-traub, line 52: Run the unit tests on a PR push.
Rather than wait for the build and then pull the artifacts, since I set up the unittests as a 'game' directory (in hopes of being able to make a very simple setup), it was more sensible to just do this since it doesn't reuse the executables in any way.

Makefile
```diff
@@ -11,7 +11,7 @@ GAME ?= standard
 
 CPLUS = g++
 CC = gcc
-CFLAGS = -g -I. -I.. -Wall -Werror -std=c++11
+CFLAGS = -g -I. -I.. -Wall -Werror -std=c++20
```
jt-traub, line 14: The new unit test framework requires C++20 support.   All of our build compilers support it.

Makefile
```diff
@@ -114,8 +120,18 @@ rules: $(GAME)/$(GAME)
 	 ./$(GAME) genrules $(GAME)_intro.html $(GAME).css html/$(GAME).html \
 	)
 
+.PHONY: unittest
+unittest:
+	$(MAKE) GAME=unittest unittest-build
+
```
jt-traub, line 126: Because the unittest needs a slightly different set of objects, it was easier just to have a new target rather than mucking with the $(GAME)/$(GAME) target to handle this oddball case.  But due to the fact that we need to game objects from *some* game in order to compile fully, having unittest be it's own (very limited) game with no ruleset/world/etc functionality made sense.

astring.cpp
```diff
@@ -389,12 +389,11 @@ ostream & operator <<(ostream & os,const AString & s)
...
-	char * buf = new char[256];
+	string buf;
 	is >> buf;
-	s.len = strlen(buf);
+	s.len = strlen(buf.c_str());
 	s.str = new char[s.len + 1];
-	strcpy(s.str,buf);
-	delete[] buf;
+	strcpy(s.str,buf.c_str());
 	return is;
```
jt-traub, line 397: These changes were necessitated by the move from C++11 to C++20.  They are semantically equivalent to what was previously there.

neworigins/map.cpp
```diff
@@ -1058,7 +1059,12 @@ void MapBuilder::SpecializeZones(size_t continents, int continentAreaFraction) {
 
 	Awrite("Grow oceans");
 	int oceanCores = std::max(1, (int) oceans.size() / 4);
-	std::random_shuffle(oceans.begin(), oceans.end());
+
+	// C++17 technically no longer has std::random_shuffle.  It's been replaced with std::shuffle and some boilerplate
+	std::random_device rd;
+	std::mt19937 g(rd());
+	std::shuffle(oceans.begin(), oceans.end(), g);
+
```
jt-traub, line 1067: The changes in this file were necessitated by the move from C++11 to C++20.  These changes are semantically equivalent to the previous code.

unit.cpp
```diff
@@ -2529,7 +2529,7 @@ int Unit::GetAttribute(char const *attrib)
 					// Ignore mage only items for non-mages
 				} else if (ap->mods[index].flags & AttribModItem::PERMAN) {
 					int men = GetMen();
-					if (men <= items.GetNum(item))
+					if (men > 0 && men <= items.GetNum(item))
```
jt-traub, line 2532: My unit tests for the unit code actually found this.   I had create a unit with no men, and then was checking what stealth levels could see it, and discovered that it was being treated as having stealth 3.   That was due to the unit having 0 men, but this code checking if the men <= amount of items (RING).. since 0 <= 0, it was true and the unit got a stealth of 3!

Making sure that men >= 0 here fixes this bug and now the stealth is 0.   I added a unit test which asserts this as part of the suite after finding it :)

artyomtrityak: oh this is an interesting discovery

jt-traub: I thought so :) j

valdisz (approved):

valdisz:
LGTM, I will push the merge button

PR #140, opened 2023-10-09
Remove the arcadia directory

jt-traub:
This code is mostly dead and hasn't been updated/built/looked at in a long time. It will still exist in git history, and I believe that Artyom is going to create 'last state/arcadia' branch which preserves it before merging all the neworigin and neworigins-v6 changes to master to bring things back to sanity.


artyomtrityak (approved):
Please also remove arcadia from a Makefile

PR #142, opened 2023-10-10
Neworigins v6 merger with master branch

artyomtrityak:
WIP, reviewing the code


jt-traub (approved):

basic/map.cpp
```diff
@@ -129,6 +129,19 @@ void ARegionList::CreateNexusLevel(int level, int xSize, int ySize, char const *
 	}
 }
 
+// void ARegionList::CreateConstrainedSurfaceLevel(int level, int xSize, int ySize, char const *name, int contients,
```
artyomtrityak, line 132: I am not sure why this is commended here

battle.h
```diff
@@ -75,11 +78,13 @@ class Battle : public AListElem
 		//
 		void UpdateShields(Army *);
 		void DoSpecialAttack( int round, Soldier *a, Army *attackers,
-				Army *def, int behind );
+				Army *def, int behind, int canattackback);
 
 		void WriteSides(ARegion *,Unit *,Unit *,AList *,AList *,int,
 				ARegionList *pRegs );
 
+		// void WriteBattleStats(ArmyStats *);
```
artyomtrityak, line 86: Need to investigate why it was commended

basic/map.cpp
jt-traub: It looks (from looking at the code and the history) that this was added by Valdis about 2 years ago for neworigins as an attempt to improve the map generation, and was added as a dummy/empty function to the other rulesets, then later removed from the game.h header and commented out everywhere.  This is (to me) prime example of code that should be just tossed since it's pretty obvious that an even better system has already replaced it.

battle.h
jt-traub: Similarly, it looks like that was your initial attempt from 3 years ago to create some battle statistics and has been replaced with newer/better code (the BattleLogLevel::Detailed stuff in the normal battle output generation).

valdisz:
This was merged as a squash merge that kills the history. It is fine when we merge changes from the _one_ author like that. But when there are many it is not correct. @artyomtrityak @jt-traub

valdisz:
Also, deletion of the Arcadia was intentional?

valdisz:
I propose to revert this PR and merge no_v6 directly into the master. This will be more correct.

jt-traub:
> Also, deletion of the Arcadia was intentional?

Yes.. I got annoyed when searching for functions that it kept going to the arcadia version rather than the real version, and since arcadia wasn't regularly being built and was quite diverged from the changes that had been made in the mainline files on neworgins, it seemed better to leave it in-state on the v5_stable branch, and get rid of it from the 'main' branch.  I am the one who authored the pr to remove it. :)

jt-traub:
> This was merged as a squash merge that kills the history. It is fine when we merge changes from the _one_ author like that. But when there are many it is not correct. @artyomtrityak @jt-traub

In general I agree with you.. I would probably have retained the branch commit history.  But I don't feel strongly enough that it needs to be reverted and remerged.  That branch history still exists in github, it just takes a bit of work to get to.

valdisz:
As I'm the one who is most annoyed with that. Then I will unwind history myself or do nothing.

PR #144, opened 2023-10-11
Generate some more complex snapshots using neworigins

jt-traub:
In order to make sure that some of the work I am doing around file output works correctly, I wanted to make sure I had some neworigins snapshots including the quest system which the standard snapshots don't have.  This checkin includes those snapshots to the point that quests have been created on turn 13.

This also updates the scripts to run them so that both sets of snapshots will be run to validate the tests.


artyomtrityak (approved):

valdisz:
More tests is good in any case

PR #145, opened 2023-10-11
Fix Landmark ordering for events

jt-traub:
The intent of the code was that weighter landmarks (settlments) would
take preference over fortifications, etc.   This test was backwards.
However, it was also possible that if there were two equally close
and equally weighty landmarks, the order would fluctate based solely on
how the in-memory layout of the unordered_map being used to construct
the data got laid out.   This changes this to be a stable ordering based
on distance (shorter distance), weight (higher weight), x coordinate
(westernmost preferred) and finally y coordinate (northernmost
preferred).   Exposed the comparison function to allow unit tests and
added unit tests.

### DIFF
--- a/events.cpp
+++ b/events.cpp
@@ -123,7 +123,10 @@ void populateSettlementLandmark(std::vector<Landmark> &landmarks, ARegion *reg, 
         .name = name,
         .title = title,
         .distance = distance,
-        .weight = 10
+        .weight = 10,
+        .x = reg->xloc,
+        .y = reg->yloc,
+        .z = reg->zloc
     });
 }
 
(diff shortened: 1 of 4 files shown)


valdisz (approved):

valdisz:
This makes sense. At first I thought that weight must be first to compare, but then I realized that you look at what is closer and then go to what is bigger :)

LGTM

jt-traub:
> This makes sense. At first I thought that weight must be first to compare, but then I realized that you look at what is closer and then go to what is bigger :)
> 
> LGTM

TBH, I could see the argument that a large city 3 hexes away from an event might be a better landmark than a hex adjacent to it, but on the whole I agree with you that closer is better followed by weight.   Future refinements might be to scale the weight based on the size of the town and the size of the fortification.  I leave that as a future improvement since it won't affect this logic, just the logic of the thing which constructs the fact landmark.

PR #146, opened 2023-10-11
Rework file input

jt-traub:
This changes the existing file input to use std::istream throughout the
bulk of the code rather than Aorderfile and Ainputfile.   Output is
still done via the custom file classes, but those will be changed
eventually. as well.

### DIFF
--- a/aregion.cpp
+++ b/aregion.cpp
@@ -103,11 +103,13 @@ TownInfo::~TownInfo()
 	if (name) delete name;
 }
 
-void TownInfo::Readin(Ainfile *f, ATL_VER &v)
+void TownInfo::Readin(istream &f)
 {
-	name = f->GetStr();
-	pop = f->GetInt();
-	hab = f->GetInt();
+	AString temp;
+	f >> ws >> temp;
+	name = new AString(temp);
+	f >> pop;
+	f >> hab;
 }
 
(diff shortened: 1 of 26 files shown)


aregion.cpp
```diff
@@ -103,11 +103,13 @@ TownInfo::~TownInfo()
 	if (name) delete name;
 }
 
-void TownInfo::Readin(Ainfile *f, ATL_VER &v)
```
jt-traub, line 106: the ATL_VER was being passed down to *some* but not all of these and never used.. I pulled it out since it wasn't needed.

astring.cpp
```diff
@@ -196,6 +196,12 @@ char *AString::Str()
 	return str;
 }
 
+
+const char *AString::const_str() const
+{
+	return str;
+}
+
```
jt-traub, line 204: We needed a way to get a const char* back out in some cases, so I added this to AString even though my intent is to get rid of it entirely at some point.

jt-traub:
These changes are what exposed the ordering issue with the global events.   When all was said and done, this code worked *except* that the global event 'location' had changed due to memory layout, and when I debugged, I saw it was choosing between 2 weight 1 options rather than using the weight 10 settlement, so dug into and fixed that bug first.

jt-traub:
Closing this PR since I need to redo it against the master branch.

PR #147, opened 2023-10-11
Rework file input

jt-traub:
This changes the existing file input to use std::istream throughout the bulk of the code rather than Aorderfile and Ainputfile. Output is still done via the custom file classes, but those will be changed eventually. as well.

### DIFF
--- a/aregion.cpp
+++ b/aregion.cpp
@@ -103,11 +103,13 @@ TownInfo::~TownInfo()
 	if (name) delete name;
 }
 
-void TownInfo::Readin(Ainfile *f, ATL_VER &v)
+void TownInfo::Readin(istream &f)
 {
-	name = f->GetStr();
-	pop = f->GetInt();
-	hab = f->GetInt();
+	AString temp;
+	f >> ws >> temp;
+	name = new AString(temp);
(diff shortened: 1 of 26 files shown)


valdisz:
As all tests and checks are passing, it means it works

aregion.cpp
```diff
@@ -103,11 +103,13 @@ TownInfo::~TownInfo()
 	if (name) delete name;
 }
 
-void TownInfo::Readin(Ainfile *f, ATL_VER &v)
+void TownInfo::Readin(istream &f)
```
valdisz, line 106: As I understand `ATL_VER &v` was not used at all...

valdisz (approved):
As the tests showed, output before and after does not change.

artyomtrityak (approved):

aregion.cpp
jt-traub: Correct.  It had been passed down to some but not all functions, and when I chased through, I found that it was not *actually* used anywhere in the reading functions *except* in the very top level to verify that the version was compatable, so I just got rid of it all the way up the call chain(s) from anything below that top level.

PR #148, opened 2023-10-15
Write all files (except rules)  using ostream

jt-traub:
This changes the code to use generic ostreams to write reports as well as game output files. 
The wrap/indent/logic which used to be part of Areport is now written as a streambuf interceptor
which inserts itself into the streams that want it.   This will allow us to (as we go farther) to use
stringstreams for testing so that we can compare/check output in test suites.

This does not *yet* modify the rules generation, nor does it *yet* support json output for
reports (and templates) but that will be coming soon.

The snapshot tests were modified (before my code changes) to generate one turn with map
output to make sure my changes did not break those formats either.

### DIFF
--- a/game.cpp
+++ b/game.cpp
@@ -131,13 +131,13 @@ AString Game::GetXtraMap(ARegion *reg,int type)
 	return(" ");
 }
 
-void Game::WriteSurfaceMap(Aoutfile *f, ARegionArray *pArr, int type)
+void Game::WriteSurfaceMap(ostream& f, ARegionArray *pArr, int type)
 {
 	ARegion *reg;
 	int yy = 0;
 	int xx = 0;
 
-	f->PutStr(AString("Map (") + xx*32 + "," + yy*16 + ")");
+	f << "Map (" << xx*32 << "," << yy*16 << ")\n";
 	for (int y=0; y < pArr->y; y+=2) {
 		AString temp;
 		int x;
@@ -147,7 +147,7 @@ void Game::WriteSurfaceMap(Aoutfile *f, ARegionArray *pArr, int type)
 			temp += GetXtraMap(reg,type);
 			temp += "  ";
 		}
-		f->PutStr(temp);
+		f << temp << "\n";
 		temp = "  ";
 		for (x=1; x< pArr->x; x+=2) {
 			reg = pArr->GetRegion(x+xx*32,y+yy*16+1);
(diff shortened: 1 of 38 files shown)


snapshot-tests/turns/turn_5/orders.3
```diff
@@ -9,6 +9,7 @@ unit 152
 ;Workers (152), revealing faction, won't cross water, 10 nomads [NOMA],
 ;  186 silver [SILV]. Weight: 100. Capacity: 0/0/150/0. Skills: none.
 @work
+option template map
```
jt-traub, line 12: Just to make sure we test the map format output, I enabled it for this turn, and then disabled it in the next turn.
I regenerated the templates (which is where these changes show up) for this turn with code from the current master.

Then coded my changes and verified that the output didn't change.

indenter.cpp
```diff
@@ -0,0 +1,51 @@
+#include "indenter.hpp"
```
jt-traub, line 1: This code, and the .hpp file are the big guts of the changes.   They are what enables me to do clean output to file
(or any iostream) which preserves the existing formatting provided by the custom file classes.

They work by acting as a buffering streambuf and insert themselves on the output streams between the  incoming data and underlying original buffer.  When data comes into the buffer, it performs the formatting required based
on the state as updated by the manipulators and then on a newline character or stream flush, output any buffered data to the underlying original buf after modification.

unittest/indent_formatter_test.cpp
```diff
@@ -0,0 +1,103 @@
+#include <sstream>
```
jt-traub, line 1: This is the other bit of net-new code, and implements unit tests (using stringstreams) to test the manipulator functionality.   Since those manipulators work on any stream, they work on stringstreams for testing and then
I know that they work correctly on filestreams which is how they get used in the actual code.

game.cpp
```diff
@@ -490,37 +481,33 @@ void Game::DummyGame()
...
+	f << PLAYERS_FIRST_LINE << "\n";
+	f << "Version: " << CURRENT_ATL_VER << "\n";
+	f << "TurnNumber: " << TurnNumber() << "\n";
 	if (gameStatus == GAME_STATUS_UNINIT)
 		return(0);
-	else if (gameStatus == GAME_STATUS_NEW)
-		f.PutStr(AString("GameStatus: New"));
+	
+	if (gameStatus == GAME_STATUS_NEW)
+		f << "GameStatus: New\n\n";
```
artyomtrityak, line 494: it is double next line here?

game.cpp
```diff
@@ -1242,18 +1219,13 @@ void Game::WriteReport()
...
-
 		if (!fac->IsNPC()) {
-			int i = f.OpenByName(str);
-			if (i != -1) {
-				fac->WriteTemplate(&f, this);
-				f.Close();
+			string str = "template." + to_string(fac->num);
+			ofstream f(str.c_str(), ios::out|ios::ate);
+			if (f.is_open()) {
+				fac->WriteTemplate(f, this);
```
artyomtrityak, line 1228: do you need to close it?

artyomtrityak (approved):

game.cpp
```diff
@@ -490,37 +481,33 @@ void Game::DummyGame()
...
+	ofstream f("players.out", ios::out|ios::ate);
+	if (!f.is_open()) return(0);
 
+	f << PLAYERS_FIRST_LINE << "\n";
+	f << "Version: " << CURRENT_ATL_VER << "\n";
+	f << "TurnNumber: " << TurnNumber() << "\n";
 	if (gameStatus == GAME_STATUS_UNINIT)
 		return(0);
-	else if (gameStatus == GAME_STATUS_NEW)
-		f.PutStr(AString("GameStatus: New"));
```
jt-traub, line 503: Just to reply to your comment that you asked and then deleted.
Yes.

The previous thing was to write the status line (which put a newline at the end of it) and then below, just write a universal newline.  I just moved that extra newline up into each of the cases since they were fixed strings.

jt-traub: Obviously verified also by the output not changing between the snapshots generated with the previous version of the code and the output from this code :)

PR #150, opened 2023-10-26
Output JSON reports - part 1

jt-traub:
Because these changes are going to be somewhat far-reaching, I am going
to do these in smaller chunks.   This chunk will contain

- a new game config which defaults to not outputting json reports.
- the inclusion of the nlohmann/json.hpp single-file json parser/manipulation library.
- Output the basic faction info (meaning the fields on the faction itself, but NOT deeper structures like regions/objects/units)
- An example of some small nested structures (skill shows/faction stats) to give examples of how this stuff will work.
- A renaming of 2 rules helper functions from link/link_ref to url/anchor in order to avoid a conflict with the link() function from unistd.h which gets pulled in via the nlohmann/json.hpp standard library includes, and which caused the rules to break.  This is JUST a renaming and no functionality change, so you can ignore all the changes in genrules.cpp as part of this.
- Removing some duplicated skill data which was being output if you were shown a skill description which allowed you create an object which in turn allowed you to use a skill (create staff of healing shows the staff of healing, which then adds a skill show for magic healing). This wouldn't affect a player since a faction tracks whether or not a skill has been shown permanently and never shows it again, but when generating the turn 1 report for the gamemaster, if those items were  enabled, 4 skills would end up shown twice.   This also accounts for  the change to the snapshot turn_0 report.1 file.  Those changes are  expected.


GAMEMASTER.md
```diff
@@ -672,7 +672,7 @@ If you wanted to switch on quartermasters, and have their level affect
 shipping costs, but not distance, you would set the variable in rules.cpp 
 like so:
 
-	Gamedefs::ALLOW_TRANSPORT | GameDefs::QM_AFFECT_COST, // TRANSPORT
+	GameDefs::ALLOW_TRANSPORT | GameDefs::QM_AFFECT_COST, // TRANSPORT
```
jt-traub, line 675: this was just a typo in the Gamemaster markdown file, so I fixed it :)

aregion.cpp
```diff
@@ -1319,6 +1319,14 @@ void ARegion::WriteExits(ostream& f, ARegionList *pRegs, int *exits_seen)
 	f << '\n';
 }
 
+void ARegion::write_json_report(json& j, Faction *fac, int month, ARegionList *pRegions) {
+	// for now, just set a name which would be the name of the region from a text report.
+	// ie "Alioth Plains (1, 2) in Alioth contains Alioth City [City]."
+	// This will get broken down and become more json-y when I flesh this code out.
+	j["name"] = Print(pRegions).const_str();
+	return;
```
jt-traub, line 1327: This is here solely so that the output shows that it *is* iterating the regions.. This is just a temp leaf.

external/nlohmann/json.hpp
jt-traub, line 1: This file is the single-include json handling library.  Like the unit test framework (ut), it requires no external libraries or links so it will just work.

faction.cpp
```diff
@@ -54,9 +53,19 @@ char const *tp[] = {
...
+// Quit states
+const string qs[] = {
+	"none",
+	"quit order",
+	"quit by gm"
+	"quit and restart",
+	"won game"
+	"game over",
+};
+const string *QuitStrs = qs;
```
jt-traub, line 67: Needed these so I could output a textual string for the various quit states in the faction json.

faction.cpp
```diff
@@ -301,144 +310,366 @@ AString Faction::FactionTypeStr()
...
+		if(citems[myfaction][i] == 0) continue;
+		size_t place = 1;
+		size_t max = 0;
+		size_t total = 0;
+
+		for (int pl = 0; pl < game->factionseq; pl++) {
+			if (citems[pl][i] > citems[myfaction][i]) place++;
+			if (max < citems[pl][i]) max = citems[pl][i];
+			total += citems[pl][i];
+		}
```
jt-traub, line 331: Since you had already iterated all items for all units for all factions while building the citems array, there was no need to do so again here, so I just rewrote this to use the data you'd already collected :p

jt-traub: Also, it now returns it as a vector so I can output it in textual and json formats as needed.

faction.cpp
```diff
@@ -301,144 +310,366 @@ AString Faction::FactionTypeStr()
...
-					((ARegionPtr*)elem)->ptr->WriteReport(f, this, pGame->month, &(pGame->regions));
-				}
-			}
-			present_regions.DeleteAll();
+	errors.DeleteAll();
+	events.DeleteAll();
+	battles.DeleteAll();
+}
+
+void Faction::write_json_gm_report(json& j, Game *game) {
```
jt-traub, line 424: the gm reports were taking up a lot of space and making the report function hard to read, so I pulled all the gm stuff out into it's own pair of functions.   I also just collect all the data up front and output it.  This made the code cleaner imho.

faction.cpp
```diff
@@ -559,34 +790,27 @@ void Faction::WriteReport(ostream& f, Game *pGame, int ** citems)
 		f << '\n';
 	}
 
-	if (shows.Num()) {
+	if (!shows.empty()) {
```
jt-traub, line 793: I converted shows/objectshows/itemsshows into std::vector<> rather than Alists..  Slowly but surely, AList and AString will die too!

faction.cpp
```diff
@@ -886,7 +1110,11 @@ void Faction::DiscoverItem(int item, int force, int full)
 		}
 	}
 	if (force) {
-		itemshows.Add(ItemDescription(item, full));
+		AString *desc = ItemDescription(item, full);
```
jt-traub, line 1113: This was leaking memory before because ItemDescription() returns a new'd AString pointer.

fracas/rules.cpp
```diff
@@ -254,6 +254,7 @@ static GameDefs g = {
 	200,	// MIN_DESTROY_POINTS,
 	34,	// MAX_DESTROY_PERCENT
 	0, // HALF_RIDING_BONUS
+	GameDefs::REPORT_FORMAT_TEXT,	// REPORT_FORMAT
```
jt-traub, line 257: This make sure noone gets the JSON format until we are ready.  I do set it in the unittest/rules.cpp so I when we eventually do end-to-end/integrations tests there as well as unit tests it will just work.

game.cpp
```diff
@@ -69,6 +72,8 @@ Game::~Game()
 	delete events;
 	ppUnits = 0;
 	maxppunits = 0;
+  	// Return the global array to it's original state. (needed for unit tests)
+  	FactionTypes->clear();
```
jt-traub, line 76: When I was testing, when I had multiple tests which set up independent games, I discovered this bug.
It's only a problem within the testing framework where you might create multiple game objects to run small tests across various files and each one would add the types configured for that test to the types array again.

game.cpp
```diff
@@ -1189,12 +1194,13 @@ void Game::WriteReport()
 	MakeFactionReportLists();
 	CountAllSpecialists();
 
-	int ** citems = new int* [factionseq];
```
jt-traub, line 1192: this memory was getting leaked previously.  Since we only need this array if we are colleting faction statistics, I cleaned it up a bit.

game.cpp
```diff
@@ -1205,19 +1211,32 @@ void Game::WriteReport()
 
 	forlist(&factions) {
 		Faction *fac = (Faction *) elem;
-		AString str = "report.";
-		str = str + fac->num;
+		string str = "report." + to_string(fac->num);
 
-		if (!fac->IsNPC() ||
-				((((month == 0) && (year == 1)) || Globals->GM_REPORT) &&
```
jt-traub, line 1212: this conditional of '(turn one || gm report) && fac == 1' was tested in a bunch of places, so I made a nice little inline function and use it here and wherever else I saw it.

game.cpp
```diff
@@ -2126,17 +2145,16 @@ void Game::CountItems (int ** citems)
 			{
 				citems[i][j] = CountItem (fac, j);
 			}
-			
-			i++;
 		}
+		i++;
```
jt-traub, line 2149: since items for npc facs get initialized to 0, leaving the npc rows in the data made it easier to avoid recomputing the factions current item totals in the rank calculator.  So now the first player faction will be the third row (assuming guard and monster npc factions) of the citems array rather than the first.  This lets me just use faction->num - 1 to reference that factions data in citems consistently.

game.h
```diff
@@ -59,6 +59,10 @@ I can see what the Item.cpp docs look like.
 class Game
 {
 	friend class Faction;
+	// Rather than making all of the functions & variables public, we can allow this special class to access them.
+	// This class won't be defined except in the unit test code, though it could be used in other places if there
+	// were a good reason to do so.
+	friend class UnitTestHelper;
```
jt-traub, line 65: this was the hack that I mentioned which allows me to do deeper unit testing where things need to get at internals of game but couldn't.  I honestly am a big believer in 'only make public the things that should be public', so will probably over time be privatizing things in other classes/structs that are only used internally, and will likely grant other UnitTestHelper friend status to those other classes if needed.

genrules.cpp
jt-traub, line 1: This entire file is just changes from link(...) -> url(....) and link_ref(...) to anchor(...).

The reason for this, as explained in the comment on the PR is that json.hpp caused the inclusion of unistd.h which defines the link() function for filesystems.   This naming conflict was easiest avoided by choosing a different name for this files internal link function.

skills.cpp
```diff
@@ -221,12 +221,6 @@ int StudyRateAdjustment(int days, int exp)
 	return rate;
 }
 
-ShowSkill::ShowSkill(int s, int l)
```
jt-traub, line 224: I made this into a struct with static initialization because it was a lot clearer.

snapshot-tests/neworigins_turns/turn_0/report.1
jt-traub, line 1: These 4 skills with just a couple ranks were caused by the enabled skills and objects such as create staff of healing, and the staff of healing.   When you learned create staff of healing, it would also add the staff of healing item if your faction hadn't already discovered it, and adding that would add the magical healing skill if your faction hadn't discovered it.

Due to the way that gm reports work (where it just produces the skill show without 'discovering' that skill or item for the faction), these ended up duplicated.  I *could* have changed the GM report to discover those items, but discovered items are stored in the game.out file for each faction and thus would have changed every snapshot turns game.out file rather than just changing the turn 0 report.1 file.   Since this only affected GM reports for the first turn this seemed reasonable.

snapshot-tests/update-all-game-snapshots.sh
jt-traub, line 1: A couple of utility shell scripts to allow updating the snapshots of the games & rules if something changes that you expect to change (such as us changing output format or something or fixing a typo).  Those would initially fail the test and then you can update all snapshots if the change was intended.

unittest/extra.cpp
```diff
@@ -36,8 +36,31 @@
 #include "gamedata.h"
 
 /// Run the initial setup for a faction
-/// which for a unit test is to do nothing
-int Game::SetupFaction( Faction *pFac ) { return 1; }
+/// For unit tests, factions get 1 leader with combat 3.
```
jt-traub, line 39: In order to unit test json reports, it still needs a valid game object correctly set up, so I construct a unit test world which has exactly 4 hexes and 1 city.  The test will also have guards and wandering monsters, so those 2 factions are taken up as well.

unittest/map.cpp
```diff
@@ -70,6 +78,7 @@ void ARegionList::MakeRegions(int level, int xSize, int ySize)
 				
 				Add(reg);
 				arr->SetRegion(x, y, reg);
+				Adot();
```
jt-traub, line 81: Just to make it easy to see how many regions got created since there are only 4 and it made testing the cout capturing have some data.

unittest/rules.cpp
```diff
@@ -168,7 +168,7 @@ static GameDefs g = {
 	0,	// UNDERWORLD_LEVELS
 	0,	// UNDERDEEP_LEVELS
 	0,	// ABYSS_LEVEL
-	100,	// TOWN_PROBABILITY
+	0,	// TOWN_PROBABILITY
```
jt-traub, line 171: I don't want it to spawn anything other than the 1 city I created.

external/nlohmann/json.hpp
jt-traub, line 1: This file is the single-include json handling library. Like the unit test framework (ut), it requires no external libraries or links so it will just work.

aregion.h
```diff
@@ -199,6 +202,7 @@ class ARegion : public AListElem
 		void WriteEconomy(ostream& f, Faction *, int);
 		void WriteExits(ostream& f, ARegionList *pRegs, int *exits_seen);
 		void WriteReport(ostream& f, Faction *fac, int month, ARegionList *pRegions);
+		void write_json_report(json& j, Faction *fac, int month, ARegionList *pRegions);
```
valdisz, line 205: Snake case detected :)
I'm not against, but that looks a bit alien around Pascal-cased statements.

valdisz:
Niels Lohmann JSON library is finally is a great addition to the codebase. This will allow many future changes.

I.e. game database files can be changed to the JSON format. It would make them less fragile and readable from the outside for different GM needs.

valdisz (approved):

faction.h
```diff
@@ -124,6 +131,27 @@ class FactionPtr : public AListElem {
...
+	string item_name;
+	size_t rank;
+	size_t max;
+	size_t total;
+
+	// This needs to be a friend function so that the json library can call it.  This can only work for simple
+	// objects/structs that require no additional parameters to be passed in.  On the other hand, for simple
+	// structures this is nice, and works with things like STL containers of structs to make the entire container
+	// serializable to json.
+	friend void to_json(json& j, const FactionStatistic& s) {
```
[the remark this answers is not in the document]
jt-traub: Yes.. (putting my reply here so future viewers of this PR if any see it).  The intent is to once all the internal work is done (which is easier when I can work on it side-by-side with the original) to move to a sort of data object (which collects just the data needed for a report from the various other classes) and the writers which take only that report data object and output it.   However, that is relatively straightforward in the future.   If c++ had reflection/introspection like a lot of other libraries, I would do something akin to that which is how most serde operations are done in other languages (kotlin, rust, etc).   Doing that in c++ requires a ton of boilerplate or macros due to the lack of reflection/introspection.

aregion.h
jt-traub: Yes...
I am highly partial to a style of TypeNames in pascal case, and functions/data elements in snake case and global constants in SCREAM_CASE.
As I rewrite pieces of this, my intent is to move more and more of it to that style, but doing a mass renaming didn't seem to make sense.

PR #151, opened 2023-10-28
Part 2 of the json reports

jt-traub:
Intent of this part is to get basic region data and the rest of  the faction data bits into the json report.

- errors, and events converted to (for now) a string vector. Will eventually make this a more interesting structure including things like unit/location when appropriate.
- Replaced the SetNPC()/IsNPC() functions with a boolean variable. There is a bit of oddity around save/load in game.in since that state was previously determined by the type values.  So added a bit of a hack to preserve data file integrity. Will likely need to live with that hack until we move to a json input/output file format for game data since converting/updating data format in current style is harder than it should be.
- Wrote out basic regions information to json output, but not yet going into buildings/units.
- Still not going into battles, but they were just text string blobs anyway in the battle 'text' member.. they should eventually become json-y objects.

### DIFF
--- a/aregion.cpp
+++ b/aregion.cpp
@@ -407,9 +407,9 @@ void ARegion::RunDecayEvent(Object *o, ARegionList *pRegs)
 	pFactions = PresentFactions();
 	forlist (pFactions) {
 		Faction *f = ((FactionPtr *) elem)->ptr;
-		f->Event(GetDecayFlavor() + *o->name + " " +
-				ObjectDefs[o->type].name + " in " +
-				ShortPrint(pRegs));
+		stringstream tmp;
+		tmp << GetDecayFlavor() << o->name << " " << ObjectDefs[o->type].name << " in " << ShortPrint(pRegs);
+		f->event(tmp.str());
 	}
 }
 
@@ -1181,7 +1181,7 @@ void ARegion::WriteProducts(ostream& f, Faction *fac, int present)
 	forlist((&products)) {
 		Production *p = ((Production *) elem);
 		if (ItemDefs[p->itemtype].type & IT_ADVANCED) {
-			if (CanMakeAdv(fac, p->itemtype) || (fac->IsNPC())) {
+			if (CanMakeAdv(fac, p->itemtype) || fac->is_npc) {
 				if (has) {
 					temp += AString(", ") + p->WriteReport();
 				} else {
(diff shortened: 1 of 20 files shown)


aregion.cpp
```diff
@@ -407,9 +407,9 @@ void ARegion::RunDecayEvent(Object *o, ARegionList *pRegs)
 	pFactions = PresentFactions();
 	forlist (pFactions) {
 		Faction *f = ((FactionPtr *) elem)->ptr;
-		f->Event(GetDecayFlavor() + *o->name + " " +
-				ObjectDefs[o->type].name + " in " +
-				ShortPrint(pRegs));
+		stringstream tmp;
+		tmp << GetDecayFlavor() << o->name << " " << ObjectDefs[o->type].name << " in " << ShortPrint(pRegs);
+		f->event(tmp.str());
```
jt-traub, line 412: As I start moving things from Astring to string, these will crop up.. they will go away again as more and more of those fields move to string from Astring since strings have good support for appending things to them but char * (which half of these are) and AStrings are half-assed at best.  So any place you see me add something like this, expect it will eventually go away.

aregion.cpp
```diff
@@ -1319,19 +1319,191 @@ void ARegion::WriteExits(ostream& f, ARegionList *pRegs, int *exits_seen)
...
+	if (type == R_NEXUS) {
+		stringstream desc;
+		desc << Globals->WORLD_NAME << " Nexus is a magical place: the entryway to the world of "
+		     << Globals->WORLD_NAME << ". Enjoy your stay; the city guards should keep you safe as long "
+			 << "as you should choose to stay. However, rumor has it that once you have left the Nexus, "
+			 << "you can never return.";
+		j["description"] = desc.str();
+	}
+
+	// hold off on writing out other stuff for now.. Just the top level region info.
```
jt-traub, line 1366: All this commented out stuff will be in the next pass.

battle.cpp
```diff
@@ -119,18 +119,8 @@ void WriteStats(Battle &battle, Army &army, StatsCategory category) {
...
-
-Battle::~Battle()
-{
-	if (asstext)
-	{
-		delete asstext;
-	}
-}
+Battle::Battle() { }
+Battle::~Battle() { }
```
jt-traub, line 123: Since I moved asstext (I didn't rename it, but should have!) from a Astring* to a std::string there is no longer any need to free.

battle.h
```diff
@@ -48,19 +51,14 @@ enum {
 	BATTLE_DRAW
 };
 
-class BattlePtr : public AListElem
-{
-	public:
-		Battle * ptr;
-};
-
```
jt-traub, line 56: the battle list on both game and factions are now std::vectors that take the Battle * directly.  No more need for this BattlePtr thingy.
Eventually everything that is an AList and derives from AListElem will be std::vector and the heavens will open up and the angels will sing hosannahs to our names.

valdisz: Vectors (or any other collection from std) is a way to go almost everywhere.

valdisz (approved):

battle.h
jt-traub: Yup.. I am converting pretty much all alists and astrings to std:: types as I go.. once I am done with json reports (and fixing the QM quandry) I will go through and do a pass and clear out the rest of the uses of those two and nuke them from orbit :p

PR #154, opened 2023-11-03
Implement the object reports

jt-traub:
This ended up being a bit larger than I was hoping.  Mostly because a bunch of the forlist() stuff annoyed me since it forced extra blocks of indentation and the lot.   So I kinda took a weed-whacker to some of it now rather than holding more of it for later passes as I had initially intended.   There are still a lot of places using Alists which I will clean up.

Verified current behavior of the new json output via new unit tests.  Verified existing behavior for text reports didn't change via the snapshot tests.

* Part 3 of json reports.   Next level down from factions to regions to objects.  This implements writing out the basic object data, but not the data related to units within the object.
* Fix a bug where unit tests were not idempotent due to the fact that the RNG was being seeded with the time the test was run.
  * To fix this , added an overriddable function pointer that the unit tests can use to ensure the same sequence of random numbers each time.
* Got rid of remaining (direct) Alists in faction. (skills and items are still using SkillList and ItemList)
* converted markets & products in aregion to std::vector


valdisz (approved):

valdisz:
What I think will be beneficial is to create helper functions that are responsible for specific parts of the JSON structure.
I.e. function that formats an item for, another function that formats coordinates, etc.

jt-traub:
> What I think will be beneficial is to create helper functions that are responsible for specific parts of the JSON structure. I.e. function that formats an item, another function that formats coordinates, etc.

Yeah.. I did some of that where pieces of data were complex/reused in some cases (see basic_json_data()) for a region which gives the info used both for the region and as part of an exit for that region, but I haven't been consistent with it.  I will do a pass back over all of this and tighten stuff up, but I have been just kinda grinding to get it done and figuring out what makes sense to refactor as I go.   I hear you though  that it would make some of this less ugly and more reusable and will get it cleaned up sooner rather than later.  Partly it's that I won't be 100% certain that the json structure is right until I try to do the final bit of 'generate the text report from the json data' which will validate that I have all the right stuff, and so I was holding off cleaning up the generative code until I was at that point, but that's an excuse :p (and I know it :) ) for a bit of laziness on my part.

stanbery:
You have no idea how happy it made me to see emails about Atlantis in the last few weeks. It's cool it's not dead yet, and it's cool that people I remember are still making it better.

jt-traub:
> You have no idea how happy it made me to see emails about Atlantis in the last few weeks. It's cool it's not dead yet, and it's cool that people I remember are still making it better.

I got sucked back into it :) :) :)

stanbery:
> I got sucked back into it :) :) :)

Well, maybe I'll try programming again one day ...

PR #158, opened 2023-11-27
Add freezing effect stage 1 in NOv6

artyomtrityak:


aregion.cpp
```diff
@@ -1564,6 +1564,12 @@ void ARegion::write_text_report(ostream &f, Faction *fac, int month, ARegionList
 			f << temp.const_str() << "\n";
 		}
 		
+		// Freezing effect
+		if (weather == W_BLIZZARD && !clearskies) {
+			temp = "There was an unnatural freezing last month.";
```
artyomtrityak, line 1569: I will need to check before turn that it does not break clients

genrules.cpp
```diff
@@ -2726,10 +2726,8 @@ int Game::GenRules(const AString &rules, const AString &css, const AString &intr
 		  << "structure which allows transportation of items.  The structures which allow this are: ";
 		last = -1;
 		comma = 0;
-		int j = 0;
```
artyomtrityak, line 2729: this has not been used, error latest compiler

monthorders.cpp
```diff
@@ -266,6 +266,12 @@ Location *Game::Do1SailOrder(ARegion *reg, Object *fleet, Unit *cap)
 			newreg = reg->neighbors[x->dir];
 		}
 		cost = 1;
+
+		// Blizzard effect
+		if (newreg && newreg->weather == W_BLIZZARD && !newreg->clearskies) {
```
artyomtrityak, line 271: this is a part of `Do1SailOrder` orders

neworigins/extra.cpp
```diff
@@ -728,6 +728,27 @@ Faction *Game::CheckVictory()
 		}
 	}
 
+	// Freezing effect
+	forlist_reuse(&regions) {
+		ARegion *r = (ARegion *) elem;
+
+		// No effect for regions with clear skies
+		if (r->clearskies) {
```
artyomtrityak, line 736: Do not activate freezing with clearskies

neworigins/extra.cpp
```diff
@@ -728,6 +728,27 @@ Faction *Game::CheckVictory()
...
+			continue;
+		}
+
+		// Check if regions in freezing zome in surface
+		if (r->zloc == ARegionArray::LEVEL_SURFACE && (r->yloc <= 0 || r->yloc >= 71)) {
+			r->Pillage();
+			r->SetWeather(W_BLIZZARD);
```
artyomtrityak, line 743: slow movement effect

monthorders.cpp
valdisz: Why do you think it is necessary to limit ships so much? What are the considerations?

artyomtrityak: The idea of freezing is to narrow down a map towards the central belt. Freezing oceans and lands is consistent and MP is 4MP for both, so horse and ship will move with the same speed

neworigins/extra.cpp
```diff
@@ -728,6 +728,27 @@ Faction *Game::CheckVictory()
...
+	forlist_reuse(&regions) {
+		ARegion *r = (ARegion *) elem;
+
+		// No effect for regions with clear skies
+		if (r->clearskies) {
+			continue;
+		}
+
+		// Check if regions in freezing zome in surface
+		if (r->zloc == ARegionArray::LEVEL_SURFACE && (r->yloc <= 0 || r->yloc >= 71)) {
```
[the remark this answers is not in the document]
jt-traub: You should mark this condition with turn > 30 which will avoid breaking the snapshot tests.  You should also write the math based on turn so it advances inward as right now you’ll need to modify the code every turn which is … suboptimal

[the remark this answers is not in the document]
jt-traub: Turn is something like (year -1 * 12 + month)-1;  I would write this as
```suggestion
                Int turn = (year - 1)*12 + month - 1;
		if (r->zloc == ARegionArray::LEVEL_SURFACE && turn > 30 && && (r->yloc <=  (turn - 30) || r->yloc >= (71 - (turn - 30))) {
```

jt-traub (commented):
As it is, it’ll break the snapshots.

neworigins/extra.cpp
```diff
@@ -728,6 +728,27 @@ Faction *Game::CheckVictory()
...
+			r->Pillage();
+			r->SetWeather(W_BLIZZARD);
+			printf("Freeze (%d,%d) region in %s of %s\n",
+				r->xloc, r->yloc, r->name->Str(), TerrainDefs[TerrainDefs[r->type].similar_type].name
+			);
+		}
+
+		// TODO: Check if regions in freezing zome in UW
```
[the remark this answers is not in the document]
jt-traub: I thought you’d previously said the UW wouldn’t freeze?

aregion.cpp
jt-traub: It likely will if any clients are parsing weather

aregion.cpp
```diff
@@ -1954,8 +1960,12 @@ int ARegion::MoveCost(int movetype, ARegion *fromRegion, int dir, AString *road)
 	int cost = 1;
 	if (Globals->WEATHER_EXISTS) {
 		cost = 2;
-		if (weather == W_BLIZZARD) return 10;
-		if (weather == W_NORMAL || clearskies) cost = 1;
+		if (weather == W_NORMAL || clearskies) {
+			cost = 1;
+		}
+	}
+	if (weather == W_BLIZZARD && !clearskies) {
```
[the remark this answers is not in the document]
jt-traub: I think this is fine, though if this code was ever part of a game that had WEATHER_EXISTS true and did something with blizzard this would break.  A game global would be safer and allow it to coexist.

genrules.cpp
jt-traub: Thanks for catching this.

neworigins/extra.cpp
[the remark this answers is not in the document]
artyomtrityak: I do not think I said this :)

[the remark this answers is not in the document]
jt-traub: Okay.. this (from the #win-conditions channel) is what led me to think it wouldn't *really* affect the underworld.

```Underworld cities do not contribute their vote (as they are safe from The Calamity (or they think so)```

I agree that it does imply that it might, but seemed to also imply it might not and I think I was just hopeful :)

PR #161, opened 2023-12-12
Write the text reports from json

jt-traub:
Rather than writing the text reports and the json reports from the internal structures, this changes the code to build the json reports once, and then use that to output both the json report (just dump the json) and the text reports (utilize the json data to output the existing textual report format).

This does change one small thing which is that now the templates are written within the WriteReports function rather than via a second pass.  Since the Json already contains the unit order info, this was the cleanest way to do it.   This requires updating the
snapshots of the turn run output since the lines about writing templates are now gone.

This looks large because it also removes all the now-unused text report writing code from everywhere else.

Note, this has been tested with snapshots, but NOT extensively under many different conditions.  It is quite possible there are bugs latent in this code due to that, and hopefully I can work with Artem between this game and the next to get some old data turns and run them with this change on the same input and validate that all the outputs match and there are no crashes, etc.


unittest/faction_test.cpp
```diff
@@ -51,18 +51,6 @@ ut::suite<"Faction"> faction_suite = []
...
-    // Mock up that a faction had set their types to 2 war and 3 magic.
-    faction->type[F_WAR] = 2;
-    faction->type[F_TRADE] = 0;
-    faction->type[F_MAGIC] = 3;
-
-    string current = faction->FactionTypeStr().Str();
-    string expected = "War 2, Magic 3";
-    expect(eq(current, expected));
-  };
-
```
jt-traub, line 65: This went away because the FactionTypeStr was only used in writing text reports, and when I got rid of those, I got rid of that function as well, so the text went the way of the dodo.

artyomtrityak (approved):

valdisz:
LGTM

PR #162, opened 2024-01-08
Change event/error to be json rather than just string arrays in json reports.

jt-traub:
Cleanup event/error to be more json-like.   This changes them to take some additional parameters, and outputs some information such as an event category and unit or region information for some events in the json.   This information is already in the text for the events, so this is just making it nicer for json-based clients in the future.

### DIFF
--- a/aregion.h
+++ b/aregion.h
@@ -183,7 +183,7 @@ class ARegion : public AListElem
 
 	public:
 		ARegion();
-		ARegion(int, int);
+		//ARegion(int, int);
 		~ARegion();
 		
 		void Setup();
@@ -197,6 +197,7 @@ class ARegion : public AListElem
 
 		int CanMakeAdv(Faction *, int);
 		int HasItem(Faction *, int);
+		json basic_region_data();
 		void build_json_report(json& j, Faction *fac, int month, ARegionList *pRegions);
 
 		AString ShortPrint(ARegionList *pRegs);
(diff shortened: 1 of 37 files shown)


aregion.h
```diff
@@ -353,6 +354,9 @@ class ARegion : public AListElem
 		int distance;
 		ARegion *next;
 
+		// A link to the region's level to make some things easier.
+		ARegionArray *level;
+
```
jt-traub, line 359: this was added here so that a region knows what level it belongs to.  This means that I don't need to pass around the entire pRegionsList everywhere, which (while it hasn't be eradicated from most places yet) will enable that to be removed from things that already know what region they are working on AND will make the code cleaner in the long run.

It was needed here since when storing a region as part of an event, I would otherwise have no way to convert that event into json without *also* storing the regions array pointer in the event, which is just bad coding style :)  And, imho, a region should know which level it belongs to, rather than just knowing the zloc and relying on passing around the entire list of arrays in order to get back to the right right.

faction.cpp
```diff
@@ -69,6 +69,23 @@ const string qs[] = {
...
 const string *QuitStrs = qs;
 
+void to_json(json &j, const FactionEvent &e) {
+	j = json{{"message", e.message}, {"category", e.category}};
+	if (e.unit != nullptr) {
+		j["unit"] = e.unit->build_json_descriptor();
+	}
+	if (e.region != nullptr) {
+		j["region"] = e.region->basic_region_data();
+	}
```
jt-traub, line 79: This is the piece of code which pretty much necessitated adding the level field to the region.

valdisz (approved):
LGTM

PR #176, opened 2024-01-24
Correctly handle destruction of the structures

valdisz:
This fixes destroy command issues.

### DIFF
--- a/game.h
+++ b/game.h
@@ -613,4 +613,7 @@ private:
 	int CountItem(Faction *, int);
 };
 
+int get_structure_health(const Object *o);
+int get_structure_points(const Object *o, int min_destroy, int max_destroy_pct);
+
 #endif
--- a/runorders.cpp
+++ b/runorders.cpp
@@ -531,6 +531,25 @@ void Game::RunDestroyOrders()
 	}
 }
 
+int get_structure_health(const Object *o) {
+	int cost = ObjectDefs[o->type].cost;
+	int health = cost - o->incomplete;
+
+	return health;
+}
+
+int get_structure_points(const Object *o, int min_destroy, int max_destroy_pct) {
+	// total cost or maxium HP of the building
+	int cost = ObjectDefs[o->type].cost;
+	int health = get_structure_health(o);
+
+	// points that can be destroyed, they are reduced by already destroyed points this turn
+	int points = std::max(0, std::max(min_destroy, cost * max_destroy_pct / 100) - o->destroyed);
+
+	// amount avaiable for destruction, but no more than there are HP of the building itself
+	return std::min(health, points);
+}
+
 void Game::Do1Destroy(ARegion *r, Object *o, Unit *u) {
 	AString quest_rewards;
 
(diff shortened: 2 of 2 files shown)


valdisz:
I'm writing unit tests for this.

runorders.cpp
```diff
@@ -550,71 +569,71 @@ void Game::Do1Destroy(ARegion *r, Object *o, Unit *u) {
...
-			o->destroyed += willDestroy;
-			o->incomplete += willDestroy;
+		default:
+			destroyPower = structurePoints;
+	}
 
-			u->Event(AString("Destroys ") + willDestroy + " structure points from the " + *(o->name) + ".");
+	int willDestroy = std::min(structurePoints, destroyPower);
+	if (willDestroy == 0) {
+		u->Error(AString("Can't destroy ") + *(o->name) + " more.");
```
jt-traub, line 606:
```suggestion
		u->error("Can't destroy " + string(o->name->const_str()) + " more.");
```

this is the new way :)

runorders.cpp
```diff
@@ -550,71 +569,71 @@ void Game::Do1Destroy(ARegion *r, Object *o, Unit *u) {
...
-			delete o;
-		}
-	} else {
-		u->Error("DESTROY: Can't destroy that.");
+	int hpLeft = std::max(0, get_structure_health(o) - willDestroy);
+	if (hpLeft > 0) {
+		o->destroyed += willDestroy;
+		o->incomplete += willDestroy;
+
+		u->Event(AString("Destroys ") + willDestroy + " structure points from the " + *(o->name) + ".");
```
jt-traub, line 619:
```suggestion
		u->event("Destroys " + to_string(willDestroy) + " structure points from the " + string(o->name->const_str()) + ".");
```

game.h
```diff
@@ -613,4 +613,7 @@ class Game
 	int CountItem(Faction *, int);
 };
 
+int get_structure_health(const Object *o);
+int get_structure_points(const Object *o, int min_destroy, int max_destroy_pct);
+
```
jt-traub, line 618: If these are only used inside runorders, there is no need to declare them here.  If you intend for them to be global, I would put them on object and then call o->get_structure_health() and o->get_structure_points(min, max).

runorders.cpp
```diff
@@ -531,6 +531,25 @@ void Game::RunDestroyOrders()
...
+	int cost = ObjectDefs[o->type].cost;
+	int health = get_structure_health(o);
+
+	// points that can be destroyed, they are reduced by already destroyed points this turn
+	int points = std::max(0, std::max(min_destroy, cost * max_destroy_pct / 100) - o->destroyed);
+
+	// amount avaiable for destruction, but no more than there are HP of the building itself
+	return std::min(health, points);
+}
+
```
jt-traub, line 552: see comment above.. if these are meant to be usable generally, make them part of Object rather than passing it in.

game.h
valdisz: First, I want them to be tested. Second, they are pure functions. I'll write tests first, and the will try to put them somewhere into better place.

jt-traub: You could have them as part of the Object class and still have them be testable.  If you need help figuring that out, let me know and I can help you.  Look at some of the stuff I did in testhelper for instance to get a Region object when I needed to test that for json export.  You could add a function there to get the Object and then call these functions on it.   I think there are similar examples in there for sending in an order and looking at results on the game state (or at least the start of it where I parse orders in so that I can see them show up in the output).. You could do something similar with a unit and a destroy order and then run the destroy order and then look at the object to see the changes for instance.

valdisz:
I will abandon this PR and instead work on the next interaction of the buildings.

PR #182, opened 2024-05-02
Fix reported bug around PHBE and illusionary dragons

jt-traub:
Looking into this also revealed that liches suffered the same problem/bug, so they have been fixed as well.   Additionally the
spell descriptions were incorrect in that they implied you could only summon one dragon or lich per turn illusorily, but the code would actually allow you to summon up to the (incorrect) limit.  Allowing this to take a number mimics the other illusory spells, and the max a unit can hold of the illusory objects mimics the max they can hold of the real objects.

Note that the `create phantasmal undead` skill is disabled, so the lich bug never showed up in a live game but it's fixed now anyway.

Additional to this, while reviewing all the summon code, I discovered that the spell description for Bird Lore 3 was incorrectly using the `.mOut` value for skeletons rather than eagles.   The actual execution code was using the correct `.mOut` value, so the correct fix was to make the spell description correct to match reality.

This does cause the snapshot turns to need to be updated since those reflected the wrong value in the spell description.

It also revealed that the most recent version of CLANG (the MacOS compiler) detected unused variables which were set and incremented but never actually read, so they were in fact dead code.  The second commit in this PR cleans that up to fix the Mac compile.

- [x] Added unit tests validating fixed behavior.

### DIFF
--- a/kingdoms/map.cpp
+++ b/kingdoms/map.cpp
@@ -1062,17 +1062,13 @@ void ARegionList::SetFractalTerrain(ARegionArray *pArr)
 
 void ARegionList::NameRegions(ARegionArray *pArr)
 {
-	int seed = 0;
-	int grown = 0;
 	Awrite("Naming Regions");
 	int unnamed = 1;
-	int toname = 0;
 	for (int x = 0; x < pArr->x; x++) {
 		for (int y = 0; y < pArr->y; y++) {
 			ARegion *r = pArr->GetRegion(x,y);
 			if (!r) continue;
 			r->wages = -1;
-			if (r->type != R_OCEAN) toname++;
 			// r->wages = AGetName(0, reg);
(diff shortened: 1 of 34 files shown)


kingdoms/map.cpp
```diff
@@ -1062,17 +1062,13 @@ void ARegionList::SetFractalTerrain(ARegionArray *pArr)
 
 void ARegionList::NameRegions(ARegionArray *pArr)
 {
-	int seed = 0;
-	int grown = 0;
```
jt-traub, line 1066: All of the variables in the function are set and incremented but never actually used, so the MacOS X compiler (CLANG) complained about them since it's static analysis code is stronger than the standard GCC code.

artyomtrityak (approved):

valdisz (approved):

PR #185, opened 2024-05-03
Prevent QM infinite chains

jt-traub:
This implements the basic protections to avoid infinitely chaining QM transports in any order.

It implements a staged transport system where first everything from all non-QM units are transported to QMs.   Then a second pass moving everything from one QM to another QM with the results ending up in a temporary holding area.   Then all QM->QM transports are resolved by putting those items from the holding area into the main items. Lastly, a final pass distributes from QM to non-QM units.

Additionally, the Transport and Distribute orders are now aliases for one another.  Both forms will work in either situation, and the phase goods move in depends on the sending unit and the target unit being QMs and if they are QMs also owning their building which must be a transport building.

Lastly, a latent bug in the PlanarDistance function where it would behave completely incorrectly if a nexus level did not actually exist has been fixed.

Additionally, the rules have been updated, a unit test has been added to validate the transport rules and a bug in the snapshot rules comparison (it was comparing two files backwards) has been fixed.


aregion.cpp
```diff
@@ -2348,8 +2348,8 @@ ARegion *ARegionList::FindNearestStartingCity(ARegion *start, int *dir)
 int ARegionList::GetPlanarDistance(ARegion *one, ARegion *two,
 		int penalty, int maxdist)
 {
-	if (one->zloc == ARegionArray::LEVEL_NEXUS ||
-			two->zloc == ARegionArray::LEVEL_NEXUS)
+	// make sure you cannot teleport into or from the nexus
+	if (Globals->NEXUS_EXISTS && (one->zloc == ARegionArray::LEVEL_NEXUS ||	two->zloc == ARegionArray::LEVEL_NEXUS))
```
jt-traub, line 2352: On a world without a nexus level, this *always* returned 10000 for planar distance :/

parseorders.cpp
```diff
@@ -638,10 +638,8 @@ void Game::ProcessOrder(int orderNum, Unit *unit, AString *o,
 			ProcessWorkOrder(unit, 0, pCheck);
 			break;
 		case O_TRANSPORT:
-			ProcessTransportOrder(unit, o, pCheck);
-			break;
 		case O_DISTRIBUTE:
-			ProcessDistributeOrder(unit, o, pCheck);
+			ProcessTransportOrder(unit, o, pCheck);
```
jt-traub, line 642: the `DISTRIBUTE` order will still be parsed (i.e., it's a valid alias for `TRANSPORT`).  There is no reason to have the commands be distinct really but the alias was left for legacy purposes.

artyomtrityak (approved):

Issue #106, opened 2022-02-10
Wrong movement rules in combat section of rules


Elfinhope:
The rules state that a unit winning a battle may not move if they have suffered any losses.  They should state that losses must be over 5% before movement is restricted.

jt-traub:
Cleaned up and fixed in PR #192

PR #192, opened 2024-05-09
Fix some additional rules issues.

jt-traub:
* Most of these came from PRs attached to Issue #106 but closed
  * Much of those PRs had been already applied, but not all.
  * Updated many instances of War/Trade to be aware of Martial
  * Fixed a long standing bug where apprentice count was duplicated.
* Updated rules to reflect the combat loss movement changes.
  * This was actually the subject of the issue and had not been fixes.
* Made the percent of loss which stops movement configurable but defaulted it in all rule sets to what it current was globally (5%)
* Update rules snapshots to reflect all the changes.

### DIFF
--- a/aregion.cpp
+++ b/aregion.cpp
@@ -3885,7 +3885,7 @@ void ARegionGraph::setInclusion(ARegionInclusionFunction includeFn) {
 	this->includeFn = includeFn;
 }
 
-void ARegionList::ResoucesStatistics() {
+void ARegionList::ResourcesStatistics() {
 	std::unordered_map<int, int> resources;
 	std::unordered_map<int, int> forSale;
 	std::unordered_map<int, int> wanted;
--- a/army.cpp
+++ b/army.cpp
@@ -1023,7 +1023,7 @@ void Army::Win(Battle * b,ItemList * spoils)
 		loses_percent = 100 - ((na * 100) / count);
 	}
 
-	if (loses_percent >= 5) wintype = WIN_DEAD;
+	if (loses_percent > 0 && loses_percent >= Globals->BATTLE_STOP_MOVE_PERCENT) wintype = WIN_DEAD;
 	else wintype = WIN_NO_DEAD;
 
 	AList units;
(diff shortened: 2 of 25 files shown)


aregion.cpp
```diff
@@ -3885,7 +3885,7 @@ void ARegionGraph::setInclusion(ARegionInclusionFunction includeFn) {
 	this->includeFn = includeFn;
 }
 
-void ARegionList::ResoucesStatistics() {
+void ARegionList::ResourcesStatistics() {
```
jt-traub, line 3888: this was just a silly naming typo, so I fixed it

army.cpp
```diff
@@ -1023,7 +1023,7 @@ void Army::Win(Battle * b,ItemList * spoils)
 		loses_percent = 100 - ((na * 100) / count);
 	}
 
-	if (loses_percent >= 5) wintype = WIN_DEAD;
+	if (loses_percent > 0 && loses_percent >= Globals->BATTLE_STOP_MOVE_PERCENT) wintype = WIN_DEAD;
```
jt-traub, line 1026: I made this a global.  The check on loses_percent is just to allow the condition to remain >= but avoid no losses = 0 when 0 is meant to mean 'any losses at all stops you'

genrules.cpp
```diff
@@ -320,8 +320,6 @@ string Game::FactionTypeDescription(Faction &fac) {
 		if (key == F_TRADE) buffer << "perform trade in " << nt << " " << plural(nt, "region", "regions");
 		if (key == F_MAGIC) {
 			buffer << "have " << nm << " " << plural(nm, "mage", "mages");
-			if (Globals->APPRENTICES_EXIST)
-				buffer << " as well as " << na << " " << Globals->APPRENTICE_NAME << (na > 1 ? "s" : "");
```
jt-traub, line 324: Believe it or not, this was actually wrong here.  There is a ruleset (Havilah) which allows 1 apprentice at Magic 0.
So in order for that rule set to show correctly, the apprentices have to be handled irrespective of if a faction is magic which is done below.
Doing it here as well resulted in all rulesets double reporting apprentices.

valdisz (approved):

PR #194, opened 2024-06-01
A couple of small bugfixes

jt-traub:
Enable JSON reports for NO v7
Fix a couple of places where the JSON->text report wasn't quite right. Fix a bug where the inactive turns handling was slightly wonky in the JSON report.


faction.cpp
```diff
@@ -409,7 +409,7 @@ void Faction::build_json_report(json& j, Game *game, size_t **citems) {
 	j["administrative"]["show_unit_attitudes"] = (showunitattitudes != 0);
 
 	if(!password_unset) j["administrative"]["password"] = password->const_str();
-	if(Globals->MAX_INACTIVE_TURNS) {
+	if(Globals->MAX_INACTIVE_TURNS != -1) {
```
jt-traub, line 412: -1 is the signifier for ignore turn inactivity, and if(-1) evaluates to true, so this was a bug but unlikely to get caught since noone actually runs with -1 as the inactive turn setup.

text_report_generator.cpp
```diff
@@ -524,9 +524,9 @@ void TextReportGenerator::output(ostream& f, const json& report, bool show_regio
 
     bool show_unit_attitudes = false;
     if (report.contains("administrative")) {
-        if (!report["administrative"].contains("times_sent"))
+        if (!report["administrative"].contains("times_sent") || !report["administrative"]["times_sent"])
```
jt-traub, line 527: I'd initially made the assumption here that the field wouldn't be output at all if it was false, but the actual generation
code always created the field and set it's value to either true or false.   For safety it will now check existence and/or the value being truthful.

In the case of times, if the value isn't set OR it's set to false then the times isn't being set.  For the password, if it's set AND it's set to true then we output the warning.

Found this while testing the new website with an downloaded report for a new faction where it had a password set..
This wasn't caught in unit tests or snapshot tests as they don't set faction passwords so they (as expected) had this warning.

valdisz (approved):

PR #195, opened 2024-06-19
Modify QM distance calculation configurably.

jt-traub:
* Added a new config value for transport to use linear connected distance (with a cross-level penalty for going through a shaft)
* Wrote a new a* search linear connected distance function.
* modified the evaluation of the order to compute the distance using either the old or new function based on the config.
* store the distance on the transport order for payment later rather than recomputing it a second time.
* Since I changed world creation to test connected distance across planes, other tests which relied on random data updated


artyomtrityak (approved):

aregion.cpp
```diff
@@ -2345,8 +2349,81 @@ ARegion *ARegionList::FindNearestStartingCity(ARegion *start, int *dir)
...
+};
+struct QEntry { ARegion *r; int dist; };
+class QEntryCompare {
+public:
+	// We want to sort by min distance
+	bool operator()(const QEntry &below, const QEntry &above) const { return above.dist < below.dist; }
+};
+
+// This doesn't really need to be on the ARegionList but, it's okay for now.
+int ARegionList::get_connected_distance(ARegion *start, ARegion *target, int penalty, int maxdist) {
```
valdisz, line 2372: Have you considered using functions in the `graphs.h` or extending that file with the new features?

valdisz: I think that A* should go to the `graphs.h` in one way or another and use stuff there.

jt-traub: I hadn't, but I will.  I looked at them this morning and they are a bit complex in some ways more than I think was needed, but I also kind of see why you did it that way.   I also need them to actually return the distance/exit early if the target isn't within the distance, so it might need another 'variant' of those.  Either way, yes, it's possible to revamp them to use it, but I might want to defer that to a followup PR/work after I get the other things I *must* have for the current game (victory condition work and map generation work).  Are you ok with deferring it?

PR #198, opened 2024-07-05
More map changes

jt-traub:
* Spread the start locations out across the world.
* Hex map now includes wing and exit info as well as gate and start location info.
* Map renderer shows start hexes and gates as well as blocked edges for the underworld.
* Work on the underworld and the map viewer
* Fix the math on corner->edge
* Update to now use the viewer options as set up by Valdis


valdisz (dismissed):
It looks good to me

artyomtrityak (approved):

PR #205, opened 2024-08-19
An iterator for AList

valdisz:
The `forlist` macro comes from when there were no iterators. `AList` is a custom container that needs to be replaced with std containers, but the amount of work is immense. Therefore I have added iterator interface (begin and end) methods so that modern `for (var item : iterator)` syntax can be used in the code. It will allow the removal of the `forlist` macro from the codebase over time.

## New API
`AList::begin` - iterator for the beginning of the list
`AList::end` - iterator for the end of the list
`AList::iter<T>` - allows for strongly typed iteration over the `AList`

## Improvements

1. Will allow the use of modern C++ `for` loop syntax.
2. All iterator-based APIs from `std` namespace will work on `AList` instances.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,23 @@
+# Editor configuration, see http://editorconfig.org
+root = true
+
+[*]
+charset = utf-8
+end_of_line = lf
+indent_style = space
+indent_size = 4
+insert_final_newline = true
+trim_trailing_whitespace = true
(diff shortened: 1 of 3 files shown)


jt-traub (approved):
This is good.  It at least makes the code cleaner in the interim until we can actually get around to removing AList.

PR #208, opened 2024-08-21
Hexside feature reimplementation from Arcadia fork

valdisz:
**This is a work-in-progress code!**

Arcadia has a very nice feature called Hexside terrain. In essence, Hexides allow additional properties to be added to the edges of the regions.

I don't think the way it was implemented in Arcadia is good enough, so I'm rolling out my own implementation.

All regions will be connected through the RegionEdge object (compared to today's direct link). This is a prerequisite for adding features to the hex edges, such as rivers, rocks, dens, roads, bridges, etc. This PR will not change anything else except how Regions are connected.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -13,9 +13,9 @@ CPLUS = g++
 CC = gcc
 CFLAGS = -g -I. -I.. -Wall -Werror -std=c++20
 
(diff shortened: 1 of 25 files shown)


valdisz:
Need to finish this one day... 😣

PR #210, opened 2024-09-17
Fix order template

jt-traub:
Fix bug where a passed-through hex showed up in order template

### DIFF
--- a/gamedata.cpp
+++ b/gamedata.cpp
@@ -4978,7 +4978,7 @@ static TerrainType td[] = {
 	 {-1,-1,-1},
 	 0,-1,-1,-1,
 	 0,{-1,-1,-1,-1,-1,-1}},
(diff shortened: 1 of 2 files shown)


gamedata.cpp
```diff
@@ -4978,7 +4978,7 @@ static TerrainType td[] = {
 	 {-1,-1,-1},
 	 0,-1,-1,-1,
 	 0,{-1,-1,-1,-1,-1,-1}},
-	{"ocean", "ocean", "barrenocean", '-', R_OCEAN,
+	{"deadwater", "deadwater", "barrenocean", '-', R_OCEAN,
```
jt-traub, line 4981: This  change is actually already in master.. it will merge cleanly.  Github is just confused because I froze the NO v7 stable branch and am merging from a branch off of that onto master.  The only real change in this commit is the change on line 900 in text_report_generator.cpp

valdisz (approved):

PR #213, opened 2024-09-18
Fix Form alias parsing misbehavior

jt-traub:
In the current game, when checking the orders, it merely checks that the alias *starts* with a number, but forms like the following

FORM 1a
FORM 1b

etc

would be allowed and would not error during the checker but would
during the run.   This makes them error during the checker phase as
well.

### DIFF
--- a/parseorders.cpp
+++ b/parseorders.cpp
@@ -2724,10 +2724,14 @@ Unit *Game::ProcessFormOrder(Unit *former, AString *o, OrdersCheck *pCheck, int 
 		return 0;
 	}
 
-	int an = t->value();
+	int an = t->strict_value();
 	delete t;
 	if (!an) {
 		parse_error(pCheck, former, 0, "Must give alias in FORM order.");
+		return 0;
+	}
+	if (an == -1) {
+		parse_error(pCheck, former, 0, "Invalid alias in FORM order.");
 		return 0;
 	}
 	if (pCheck) {


sgb:
Internally "FORM 1a ... FORM 1b" is currently equivalent to "FORM 1 ... FORM 1", and the checker has never reported alias conflicts.  However "FORM 1a ... FORM 2b" WOULD have worked, and the checker now disallows this.

Since the rules do state that the alias is a number this could be considered user error, and it isn't a bad thing to improve the checker - but it seems the underlying issue is not allowing alphanumeric alias suffixes, but the checker's alias clash blindness.

As such, perhaps a better fix would be to check for alias clashes and keep ignoring alpha suffixes.  This is obviously more complex, but would be fixing the actual problem (and guarding against some other very common mistakes).

valdisz (approved):

jt-traub:
> Internally "FORM 1a ... FORM 1b" is currently equivalent to "FORM 1 ... FORM 1", and the checker has never reported alias conflicts. However "FORM 1a ... FORM 2b" WOULD have worked, and the checker now disallows this.
> 
> Since the rules do state that the alias is a number this could be considered user error, and it isn't a bad thing to improve the checker - but it seems the underlying issue is not allowing alphanumeric alias suffixes, but the checker's alias clash blindness.
> 
> As such, perhaps a better fix would be to check for alias clashes and keep ignoring alpha suffixes. This is obviously more complex, but would be fixing the actual problem (and guarding against some other very common mistakes).

I don't *disagree* with you, but I think this is still better than the current situation.. the 1a and 2b would however have been parsed as 1 and 2 internally and the alpha suffix ignored.  Since the form order really does require a *number* not just an alias, this seemed like a reasonable enforcement of an existing rule and reporting it on error check seemed reasonable.   I agree that checking for alias conflicts would be helpful, but that makes the checker stateful which is isn't today and I don't want to add stateful-ness there as of right now.

PR #216, opened 2024-09-26
Clean up all uses of AList from lists of Units

jt-traub:
This makes units (one of the largest things which used alists) no longer use them.. This is one step on the road toward more modern stl containers and patterns being the way this code works.


valdisz (approved):
Less `forlist`s is better :)

jt-traub:
> Less `forlist`s is better :)

End goal is 0 `forlist`s and 0 `AListElem` and 0 `AList` :)

PR #217, opened 2024-09-27
Remove 2 more uses of ALists

jt-traub:
This removes the use of ALists of strings (in oldorders and turnorders), as well as the use of ALists for Skills and SkillList.
SkillList was reimplemented as a class containing a private std::list container and exposing the iterators and the needed
functions.


valdisz (approved):
LGTM

PR #218, opened 2024-09-28
Remove more uses of AList/AListElem

jt-traub:
* Shields are now an stl container.  Reimplemented get highest shield in terms of the STL in-place since it was only used once and got rid of the shieldlist class as well.  In order to simplify memory management replaces the Shield * with shared_ptr<Shield> which provides a smart ref-counted pointer, so that just clearing (or erasing) an element from the container will delete the allocated memory.  Moved the definition of Shield into army.h which was the only thing that used it now.
* Removed all uses of AList from the quests system (quests and rewards). Retained the QuestList class since it had a number of useful functions and reworked it using an internal std::vector.  Renamed all the member functions to follow intended style going forward.
* Removed vestiges of AList from productions. This had mostly been done, so this was just cleaning up a leftover place or two where it hadn't been fully removed.  Also renamed the production class's functions using the new style.
* Removed vestiges of AList from markets. This has been partially done previously, this was just cleaning up some leftover turds.  Renamed Market class functions to use new style of naming.  Made the enum type for the market into a named enum and part of the Market class for scoping reasons.


market.cpp
```diff
@@ -28,26 +28,7 @@
 #include "gameio.h"
 #include "gamedata.h"
 
-Market::Market()
```
jt-traub, line 31: I moved these into the header file with the more modern style of member variable initialization.

market.cpp
```diff
@@ -81,18 +61,14 @@ void Market::PostTurn(int population, int wages)
 
 	if (population <= minpop)
 		amount = minamt;
+	else if (population >= maxpop)
+		amount = maxamt;
 	else {
-		if (population >= maxpop)
```
jt-traub, line 85: Just denesting some ifs.

market.h
```diff
@@ -56,22 +61,8 @@ class Market : public AListElem {
 	int baseprice;
 	int activity;
 
-	void PostTurn(int, int);
-	void Writeout(ostream& f);
-	void Readin(istream& f);
-	//AString Report();
-};
-
-class MarketList : std::vector<Market*> {
```
jt-traub, line 65: This wasn't used, *AND* deriving from STL containers is a bad idea since they don't have virtual destructors.

valdisz (approved):
lgtm

PR #221, opened 2024-09-29
Update entity weight

jt-traub:
The intent was that the entities are not transportable via magic, via transport or via the underworld, so that they cannot be moved around except via overland/ships. However, I had mistakenly set their weight such that high level portal lore and teleport could be used to move them. The easiest fix was to up their weight so that this becomes impossible, though the side effect is they now require more ships to transport. This shouldn't be a problem however in the scheme of things as the larger ships can handle them in a reasonable number and even the smaller ships can do it, with enough of them in the fleet.


valdisz:
just for the master branch?

jt-traub:
> just for the master branch?

Yeah.. I don't need PR approvals to merge into the NO7 branch, so I did that already :)

PR #222, opened 2024-10-07
Remove AList from use in orders/order lists

jt-traub:
* Actually do a named enum correctly in order to enforce typechecking. Also, modified editor settings to get rid of dumb extra spaces, and wrap things better, so these will occur over time as I edit files.
* Modifies all orders to no longer be ALists.  Also moves them to be shared_ptr rather than using raw pointers with
new/delete. In most cases, I chose to use std::vector for the list of orders but in some (two) cases I chose std::list since there was a need to insert at the head.  Choosing vector over list is preferred in general as it has better speed, but it makes inserting at the head harder.


valdisz (approved):

valdisz:
For a non-C++ dev... What purpose does `shared_ptr` have?

jt-traub:
> For a non-C++ dev... What purpose does `shared_ptr` have?

It's a smart pointer.  The underlying memory is only freed when all references to it go away.  For a local variable (like one within a function) that happens when the scope closes.  For one in a object (or a container) it happens when a null pointer is assigned into it, or when the parent object is destructed or when the container is deleted or the object is erased from the container.

If you have two variables pointing to the same location, you have a reference count of 2 so the lifetime of the underlying memory is the maximum of the lifetime of all the variables referencing it.

Smart pointers incur a bit of memory overhead, but they are safer, and they also mean you don't need to manually call delete everywhere, especially in local code, and thus you can get rid  of a lot of special-case cleanup code in most places, making for better readability.

unique_ptr is a similar smart pointer but guarantees one and only one owner of the memory at a time (i. e., you cannot create another pointer to the same location).   weak_ptr is a similar concept and works with shared_ptr actually, to provide a weak reference (ie, the object knows if there is a weak reference to it, but it doesn't increase the reference count, so the object can be removed when the reference count is reduced to 0 and the weak pointer object will know that it's associated shared pointer was deleted and can be checked before use -  when I start converting other pointers to shared_ptr you will likely see some weak ptr references for things like the units link to it's containing hex, or other similar backreferences.

PR #236, opened 2025-01-25
Fix bug with assassinating STED/CATA

jt-traub:
If a unit contained only STED/CATA it would crash the game if someone attempted to assassinate it.

While it's possibly non-ideal to allow assassinating inanimate things, the way the code is currently structured it's a valid target unit.

This change makes it work, AND will prefer to assassinate a man over an automaton in a mixed unit.


valdisz (approved):
LGTM

Should be mentioned on the homepage that you CAN assassinate sted or catp as current knowledge of the players is that you can't do it.

PR #237, opened 2025-01-28
Make portal behave the same as teleport

jt-traub:
When portals were added, it's range wasn't adjusted under the easier underworld settings like all of the rest.  This fixes that.


valdisz (approved):
LGTM :)

PR #268, opened 2025-05-24
Remove some more AString use

jt-traub:
Clean up a few functions which return AString and/or AString * or using AString & arguments.

Also included are some newer cursor rules that I have been trying out, but those can be ignored for the review as they only affect my use of cursor dev environment :)


valdisz:
Do we really need that Cursor related stuff there?

valdisz:
> > Do we really need that Cursor related stuff there?
> 
> Not really, but I’d asked a while back if you all wanted me to exclude it completely and y’all said no, lol.

My memory is short :D

jt-traub:
> I still don't like Cursor....

Fair, and I will remove all the cursor rules in the next PR at this point.  1) the rules aren't doing what I had really hoped they would to help clean up the code more quickly and 2) as they update they generate large diffs which aren't relevant to most :)

PR #272, opened 2025-07-19
Fix several out-of-bounds array reads

ennorehling:


basic/map.cpp
```diff
@@ -822,9 +822,9 @@ void ARegionList::MakeUWMaze(ARegionArray *pArr)
             ARegion *reg = pArr->GetRegion(x, y);
             if (!reg) continue;
 
-            for (int i=D_NORTH; i<= NDIRS; i++) {
+            for (int i=D_NORTH; i != NDIRS; i++) {
                 int count = 0;
-                for (int j=D_NORTH; j< NDIRS; j++)
+                for (int j=D_NORTH; j != NDIRS; j++)
                     if (reg->neighbors[j]) count++;
```
jt-traub, line 828: I actually don't see the logical point to make this changes.  I *believe* that they are logically effectively equivalent and so the change feels stylistic rather than functional and while I do make stylistic changes at times, I don't particularly see the point of this one.

jt-traub: If you can show me the case where this would cause some out of bounds error, I would be more willing to accept it, but I do not see it.

economy.cpp
```diff
@@ -1099,14 +1099,16 @@ int ARegion::TownHabitat()
...
         if (ObjectDefs[obj->type].protect > fort) fort = ObjectDefs[obj->type].protect;
-        if (ItemDefs[ObjectDefs[obj->type].productionAided].type & IT_FOOD) farm++;
-        if (ObjectDefs[obj->type].productionAided == I_SILVER) inn++;
-        if (ObjectDefs[obj->type].productionAided == I_HERBS) temple++;
-        if (
-            (ObjectDefs[obj->type].flags & ObjectType::TRANSPORT) &&
-            (ItemDefs[ObjectDefs[obj->type].productionAided].type & IT_MOUNT)
-        ) {
-            caravan++;
+        if (ObjectDefs[obj->type].productionAided >= 0) {
```
jt-traub, line 1102: This makes sense, since I believe you are correct that productionAided can in fact be initialized to -1 which is somewhat common for a lot of integral values in Atlantis (another thing I want to eventually clean up as I find that style abhorrent).  Regardless, this change makes sense to me.

fracas/map.cpp
```diff
@@ -823,9 +823,9 @@ void ARegionList::MakeUWMaze(ARegionArray *pArr)
             ARegion *reg = pArr->GetRegion(x, y);
             if (!reg) continue;
 
-            for (int i=D_NORTH; i<= NDIRS; i++) {
```
ennorehling, line 826: Here, `i` will be NDIRS in the last iteration of the loop, and access reg->neighbors[NDIRS], which is one past the end of the array. By breaking the loop once i reaching NDIR, the bug is avoided.

jt-traub:
This is valid.. there was in fact a potential out of bounds during initial UW maze creation on the outer loop which incorrectly did i <= NDIRS rather than i < NDIRS.

PR #273, opened 2025-07-19
remove dynamically sized arrays.

ennorehling:
These are a C99 feature not in the C++ standard (although GCC implements them, other compilers do not).
For now, let's use a std::vector instead.

### DIFF
--- a/aregion.cpp
+++ b/aregion.cpp
@@ -2570,12 +2570,9 @@ void makeRivers(
     logger::write("Distances from water body to water body");
 
     size_t sz = waterBodies.size();
-    int distances[sz][sz];
-    for (size_t i = 0; i < sz; i++) {
-        for (size_t j = 0; j < sz; j++) {
-            distances[i][j] = INT32_MAX;
-        }
-    }
+    std::vector<int> distances;
+    distances.resize(sz * sz);
+    std::fill(distances.begin(), distances.end(), INT_MAX);
 
     for (const auto& water : waterBodies) {
         logger::write("WATER BODY " + std::to_string(water->name));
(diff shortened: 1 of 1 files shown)


jt-traub:
This is something I had been doing as I found them.  I have been replacing *most* static (and dynamic) arrays either with std::array or std::vector, so this will get merged soon-ish (basically I want to validate that changing this doesn't modify the output of a 'real' game -- I have been using v7 mid-game turns for this, which has helped me avoid many potentially breaking changes to behavior as I have been working.

jt-traub:
> I agree that this could use a std::array instead of a vector. The goal here is to hold the memory and have it be de-allocated at the end of scope, which the vector does, too, but it comes with the whole container behavior which this code doesn't need. std::array makes it more clear what we're trying to do.
> 
> I'll push an update ASAP!
> 
> Edit: Oh no, I'm showing my inexperience here. I've never used std::array before, I didn't realize that you declare it with a fixed size. If the size is not constant like here, where it;s `sz * sz`, then std::vector _is_ the way to do it. Any other objections to my change?
> 
> What I really wanted here is an auto_ptr to a new int *, but auto_ptr is gone.

Oh, I wasn't saying that this should be std::array instead of std::vector.  I was saying that I have been fixing dynamically and statically allocated c-style arrays with those when I spot them and have the time to clean them up, so this was an appreciated change.

jt-traub:
Having take another look at this, I do like removing the dynamically sized C array, but I feel that the code was more readable actually using a 2d array rather than the conversion you did to a single-dimensioned array.. I will actually be reworking this.  I will tag you on my redone PR and then will have Valdis or Artem approve it.

PR #274, opened 2025-07-19
GCC 11 doesn't support ranges properly

ennorehling:
On Ubuntu 22.04, the default GNU C compiler version is gcc-11. It doesn't understand the code in string_filters.cpp, because the ranges library wasn't fully implemented at the time.

It's possible to achieve the same result with a slightly less efficient implementation, which I'm only enabling if required.

It would of course be nicer to detect the compiler and set -DWITHOUT_VIEWS in the Makefile, not in string_filters.hpp, but I don't know how to do that.

### DIFF
--- a/string_filters.hpp
+++ b/string_filters.hpp
@@ -128,15 +128,14 @@ inline constexpr lowercase_t lowercase{};
  */
 struct canonicalize_t {
     std::string process(const std::string& str) const {
+        std::vector<std::string> processed_words;
 #ifdef CPP20_RANGES_ARE_BROKEN
         /** gcc 11 doesn't like the range solution, ergo this hack */
-        std::vector<std::string> words;
         std::istringstream f(str);
         std::string s;
         while (std::getline(f, s, ' ')) {
-            words.push_back(s | capitalize);
+            processed_words.push_back(s | capitalize);
         }
-        return strings::join(words, "_");
 #else
         auto parts_view = str
             | std::views::split(std::string_view{" "}) // Split by space
@@ -145,12 +144,11 @@ struct canonicalize_t {
                 return str | capitalize;
             });
 
-        std::vector<std::string> processed_words;
         // Efficiently copy the view elements into the vector
         std::ranges::copy(parts_view, std::back_inserter(processed_words));
+#endif
         // Join the processed words with an underscore
         return strings::join(processed_words, "_");
-#endif
     }
 
     std::string operator()(const std::string& str) const { return process(str); }
--- a/string_filters.hpp
+++ b/string_filters.hpp
@@ -149,11 +149,9 @@ struct canonicalize_t {
             if (!s.empty()) {
                 s.append(1, '_');
             }
-            // s += w | capitalize;
-            s.append(1, (char) std::toupper(w[0]));
-            s.append(w.begin() + 1, w.end());
+            s += w | capitalize;
         }
-        return str; // WRONG!
+        return str;
 #else
         auto parts_view = str
             | std::views::split(std::string_view{" "}) // Split by space
--- a/string_filters.hpp
+++ b/string_filters.hpp
@@ -1,7 +1,14 @@
 #pragma once
 #ifndef STRING_FILTERS_HPP
 #define STRING_FILTERS_HPP
-
+#ifndef WITHOUT_VIEWS
+#ifdef __GNUC__
+#if __GNUC_PREREQ(12,0)
+#else
+#define WITHOUT_VIEWS
+#endif
+#endif
+#endif
 #include "strings_util.hpp"
 
 #include <algorithm>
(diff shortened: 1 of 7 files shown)


jt-traub:
As I said on the game channel, I would prefer to support only libc++20 or newer robustly, which is pretty much g++12 or newer.  I appreciate that this would make it more available to a number of people (and yes, the default g++ on 22.04 is gcc-11, but gcc-12 is available and officially supported and (supposedly since I haven't tested it) fully supports the ranges library).  If that does work, I would prefer not to support older version than that at this time.

string_parser.hpp
```diff
@@ -55,7 +54,7 @@ std::istream& operator>>(std::istream& is, string_parser& parser);
 
 class string_parser {
 public:
-    constexpr string_parser() noexcept : data_{}, pos_{0} {}
+    string_parser() noexcept : data_{}, pos_{0} {}
```
jt-traub, line 57: Am curious on the reason for removal of the constexpr here?

string_filters.hpp
```diff
@@ -124,6 +133,23 @@ inline constexpr lowercase_t lowercase{};
...
+            words.push_back(s);
+        }
+        s.clear();
+        for(const std::string& w : words) {
+            if (!s.empty()) {
+                s.append(1, '_');
+            }
+            s += w | capitalize;
+        }
+        return str;
```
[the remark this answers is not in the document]
jt-traub: in fact, I believe this should return strings::join(words, "_");

I would need to verify this using that version of gcc to be sure this is correct however.

string_filters.hpp
```diff
@@ -124,6 +133,23 @@ inline constexpr lowercase_t lowercase{};
  */
 struct canonicalize_t {
     std::string process(const std::string& str) const {
+#ifdef WITHOUT_VIEWS
+        /** FIXME: gcc 11 doesn't like the range solution, ergo this hack */
+        std::vector<std::string> words;
+        std::istringstream f(str);
+        std::string s;
+        while (std::getline(f, s, ' ')) {
+            words.push_back(s);
```
jt-traub, line 142: To go with the return suggestion I made below, this should probably be words.push_back(s | capitalize)

string_filters.hpp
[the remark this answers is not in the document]
ennorehling: I think you're right, this should return s. These filter functions could have their own unit tests, I suppose?

ennorehling:
I added a test and fixed the non-ranges canonicalize implementation. Also decided to rename the preprocessor flag that enables it, and removed the compiler version checks, to be set by Makefile or CMake instead.

string_filters.hpp
```diff
@@ -124,6 +128,16 @@ inline constexpr lowercase_t lowercase{};
...
+#ifdef CPP20_RANGES_ARE_BROKEN
+        /** gcc 11 doesn't like the range solution, ergo this hack */
+        std::vector<std::string> words;
+        std::istringstream f(str);
+        std::string s;
+        while (std::getline(f, s, ' ')) {
+            words.push_back(s | capitalize);
+        }
+        return strings::join(words, "_");
+#else
```
jt-traub, line 140: This could actually be tightened up further I believe.
```suggestion
        std::vector<std::string> processed_words;
#ifdef CPP20_RANGES_ARE_BROKEN
        /** gcc 11 doesn't like the range solution, ergo this hack */
        std::istringstream f(str);
        std::string s;
        while (std::getline(f, s, ' ')) {
            processed_words.push_back(s | capitalize);
        }
#else
```

jt-traub: Then in the else block, remove the definition of processed_words, and the return, and then

string_filters.hpp
```diff
@@ -136,6 +150,7 @@ struct canonicalize_t {
         std::ranges::copy(parts_view, std::back_inserter(processed_words));
         // Join the processed words with an underscore
         return strings::join(processed_words, "_");
+#endif
```
jt-traub, line 153:
```suggestion
#endif
        return strings::join(processed_words, "_");
```

Doing it this way just changes the way processed_words are built and keeps the rest of the logic uniform, which is clearer intent imho.

jt-traub (commented):
Thanks for adding those unit tests.  If we can get this branch rebased onto current master to resolve the merge conflicts, then I can approve it.

string_parser.hpp
ennorehling: I had some compilation problems with them, don't remember from which compiler. Removing the constexpr made the code build, and I never thought about it again. I'm not incredibly versed in the correct use of this keyword, but if you want, I'll find out what my issue was, and if there's a better solution.

jt-traub: If so, we might need to figure something out.

PR #275, opened 2025-07-19
reading gamenames that contain whitespace

ennorehling:
"Standard Atlantis" was causing problems.
Read the entire line, instead of just the first word.

### DIFF
--- a/game.cpp
+++ b/game.cpp
@@ -393,7 +393,8 @@ int Game::OpenGame()
     }
 
     std::string gameName;
-    f >> std::ws >> gameName;
+    f >> std::ws;
+    std::getline(f, gameName);
     if (f.eof()) return(0);
 
     if (!(gameName == Globals->RULESET_NAME)) {


game.cpp
```diff
@@ -393,7 +393,8 @@ int Game::OpenGame()
     }
 
     std::string gameName;
-    f >> std::ws >> gameName;
+    f >> std::ws;
+    std::getline(f, gameName);
```
jt-traub, line 397: This is definitely a bug, and I am somewhat surprised that I didn't catch this one in testing.. thank you.

jt-traub (approved):

game.cpp
ennorehling: one can always write more automated tests!

jt-traub: > one can always write more automated tests!

Yes, though testing the file input/output handling in the unit tests is actually harder, and for that I tend to rely on the smoke tests or the (not checked in since they used real-game data) regression tests I have been running with the v7 game turns.  The snapshot tests missed it because the game files in there for some reason although they use the 'Standard' binary,  which was the 'Wyreth' game rules from 4.10.x.. the only one which has the space is the 'basic' engine (and the unit test game which doesn't read/write the files so it was never spotted).. anyway.. yes, this should be tested for and I will find a way :)

Issue #248, opened 2025-03-28
Remove uses of AString


jt-traub:
Code quality improvement.

jt-traub:
Split this off from the AList/AString one as each is large independent of the other so it makes sense to tackle them seperately.

jt-traub:
Thanks to work by @ennorehling  this is now done.  There are still uses of char * that I want to get rid of but astring in all it's abhorrent glory is gone.

PR #276, opened 2025-07-20
Remove AString

ennorehling:
Just use std::string everywhere.


ennorehling:
This should resolve #248

jt-traub (approved):

PR #277, opened 2025-07-20
Add CMake project definition

ennorehling:
[CMake](https://cmake.org) is much easier to maintain than hand-crafted Makefiles, and has IDE support in Visual Studio and VS Code. This change doesn't replace the existing Makefile builds, but adds CMake as an option.

Accepting this PR would make life a lot easier for me!

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,0 +1,51 @@
+cmake_minimum_required(VERSION 3.31)
+project(atlantis CXX)
+set(CMAKE_CXX_STANDARD 20)
+
+add_library(engine
+    aregion.cpp
+    army.cpp
+    astring.cpp
+    battle.cpp
+    economy.cpp
+    edit.cpp
+    faction.cpp
+    game.cpp
+    gamedata.cpp
+    gamedefs.cpp
+    genrules.cpp
+    items.cpp
+    main.cpp
+    market.cpp
+    modify.cpp
+    monthorders.cpp
+    npc.cpp
+    object.cpp
+    orders.cpp
+    parseorders.cpp
+    production.cpp
+    quests.cpp
+    runorders.cpp
+    skills.cpp
+    skillshows.cpp
+    specials.cpp
+    spells.cpp
+    unit.cpp
+    events.cpp
+    events-battle.cpp
+    events-assassination.cpp
+    mapgen.cpp
+    simplex.cpp
+    namegen.cpp
+    indenter.cpp
+    text_report_generator.cpp
+)
+add_executable(basic
+    basic/extra.cpp
+    basic/map.cpp
+    basic/monsters.cpp
+    basic/rules.cpp
+    basic/world.cpp
+)
+target_link_libraries(basic engine)
+target_include_directories(basic PUBLIC ${CMAKE_CURRENT_SOURCE_DIR})
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,6 +1,20 @@
-cmake_minimum_required(VERSION 3.31)
+cmake_minimum_required(VERSION 3.22)
 project(atlantis CXX)
 set(CMAKE_CXX_STANDARD 20)
+
+if (CMAKE_COMPILER_IS_GNUCC OR CMAKE_C_COMPILER_ID MATCHES "Clang")
+    # add_compile_options(-Wsign-compare -Wall -Werror -Wno-switch -Wno-unknown-pragmas -Wstrict-prototypes -Wpointer-arith -Wno-char-subscripts -Wno-long-long)
+    add_compile_options(-Wall -Werror -pedantic)
+elseif(MSVC)
+    add_compile_options(/WX /MP /FC)
+    set(CMAKE_EXE_LINKER_FLAGS_DEBUG
+        "${CMAKE_EXE_LINKER_FLAGS_DEBUG} /NODEFAULTLIB:libc.lib /NODEFAULTLIB:libcmt.lib /NODEFAULTLIB:libcd.lib /NODEFAULTLIB:libcmtd.lib /NODEFAULTLIB:msvcrt.lib")
+    set(CMAKE_EXE_LINKER_FLAGS_RELEASE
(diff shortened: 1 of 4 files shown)


jt-traub:
As I said on the discord, I am a bit leery of this.  However, I can see the rationale.  What I don't want however is to maintain 2 build systems.  I will investigate whether or not this is feasible to switch to entirely, but ultimately, that decision will be up to Artem.  I am curious as to your choice of warnings rather than -Wall -Werr however, shouldn't the latter be *more* than what you are covering?  If not, then the strongest set of error/warning conditions should be what is in both places which is what I want in order to make sure we are not introducing errors as we go.

CMakeLists.txt
```diff
@@ -0,0 +1,100 @@
...
+add_executable(${GAME}
+    main.cpp
+    ${GAME}/extra.cpp
+    ${GAME}/map.cpp
+    ${GAME}/monsters.cpp
+    ${GAME}/rules.cpp
+    ${GAME}/world.cpp
+)
+target_link_libraries(${GAME} engine)
+target_include_directories(${GAME} PUBLIC ${CMAKE_CURRENT_SOURCE_DIR})
```
jt-traub, line 74:
```suggestion
    ${GAME}/world.cpp
    $<TARGET_OBJECTS:engine>
)
target_include_directories(${GAME} PUBLIC ${CMAKE_CURRENT_SOURCE_DIR})
```

CMakeLists.txt
```diff
@@ -0,0 +1,100 @@
...
+    unittest/extra.cpp
+    unittest/map.cpp
+    unittest/monsters.cpp
+    unittest/rules.cpp
+    unittest/world.cpp
+    unittest/testhelper.cpp
+    ${TEST_SOURCES}
+)
+target_link_libraries(unittest engine)
+target_include_directories(unittest PUBLIC ${CMAKE_CURRENT_SOURCE_DIR})
```
jt-traub, line 97:
```suggestion
    unittest/testhelper.cpp
    ${TEST_SOURCES}
    $<TARGET_OBJECTS:engine>
)
target_include_directories(unittest PUBLIC ${CMAKE_CURRENT_SOURCE_DIR})
```

jt-traub:
Looking at it a bit, it seems like it would also be better to do a *real* test for whether ranges are correctly supported (which will help resolve the other PR around that as well)

```
include(CheckCXXSourceCompiles)

check_cxx_source_compiles("
#include <ranges>
#include <vector>
#include <algorithm>

int main() {
    std::vector<int> v = {1, 2, 3, 4, 5};
    auto squared = v | std::views::transform([](int x) { return x * x; });
    std::vector<int> result;
    std::ranges::copy(squared, std::back_inserter(result));
    return 0;
}
" HAS_RANGES)

if (!HAS_RANGES)
    message(STATUS "C++20 Ranges library not detected.")
    add_compile_options(-DHAVE_BROKEN_RANGES)
endif()
```

Seems to be valid cmake magic to have the compiler tell us if it supports them correctly.  If you could give that (and the object library changes above) a whirl on your machine and let me know if they work, maybe we can get this all merged in for you.

ennorehling:
Still missing the "rules" target from the original Makefile, going to add that before removing the Draft flag from this PR.

main.cpp
```diff
@@ -38,7 +38,7 @@ int main(int argc, char *argv[])
     // For simplicity, convert all arguments to strings
     std::vector<std::string> args;
     for (int i = 0; i < argc; i++) {
-        args.push_back(argv[i] | filter::lowercase);
+        args.push_back(argv[i]);
```
jt-traub, line 41: why this unrelated change?

CompilerSupport.cmake
```diff
@@ -0,0 +1,26 @@
...
+        | std::views::transform([](auto&& subrange) {
+            return std::string(subrange.begin(), subrange.end());
+        });
+
+    std::vector<std::string> processed_words;
+    std::ranges::copy(parts_view, std::back_inserter(processed_words));
+    return (2 == processed_words.size()) ? 0 : 1;
+}
+" COMPILER_HAS_RANGES)
+endif()
```
jt-traub, line 26: Why not just do the compile test in all cases, that way if a future msvc compiler somehow breaks range support we catch it here.  Doing this compile check is pretty damned snappy and just doing it always feels better than those explicit checks.

main.cpp
ennorehling: Because t was wrong. For example, my code is checked out in a directory named Atlantis, and lower-casing the full path to the intro.html file will cause it to not be found by the genrules command.

CompilerSupport.cmake
ennorehling: it isn't *that* snappy, and chances of a future compiler version breaking something that works are very low. You have to be able to trust some fundamental stuff.

main.cpp
jt-traub: Hmmm yeah.. a better choice there would be to use `std::vector<strings::ci_string>` which preserves the original case of the input, but compares arguments case insensitively, which mirrors the original AString behavior (this change here predates my introduction of that ci_string class for those specific cases which needed it).. I am ok with this, and will probably just change it to use ci_string sometime in the (near) future unless you want to do it as part of this?

CompilerSupport.cmake
jt-traub: I suppose, I just tend to have an aversion to testing things with magic numbers (ie, 143 and 11). [and yes, I know the code is *rife* with those, and I want to get rid of them all eventually]

main.cpp
jt-traub: strings::ci_string doesn't convert to lowercase, just applies case-insenstive comparators.  As I said, this was (as you pointed out) a broken attempt to retain the original AString comparison behavior when I removed AString from the file and argument handling initially.  This code *used* to convert all arguments to AStrings. strings::ci_string was the cleaned up version of that but I forgot to fix it here.  Either way, I am willing to accept this piece as is.

jt-traub (approved):

PR #279, opened 2025-07-22
unittests create 2 levels, but only one was allocated

ennorehling:


unittest/world.cpp
```diff
@@ -9,7 +9,7 @@ const std::string& AGetNameString(int name) {
 
 void Game::CreateWorld() {
     logger::write("Creating world");
-    regions.create_levels(1);
+    regions.create_levels(2);
```
jt-traub, line 12: Ahh. yes, this is true.. now.. why the hell did this cause a unit test failure... I will need/want to figure that out since that's concerning why the unit test is now failing, but only the 1 level is being allocated.. which means I am likely stomping on memory somewhere in the unit tests which is quite confusing that they weren't failing before.

PR #280, opened 2025-07-23
Fix unit test small bugs.

jt-traub:
This subsumes and includes Enno's changes from PR #279 and on top of that fixes some other small nits.

1) it removes some unneeded debug code that had been left around.
2) it fixes the 2 unit tests which changed due to fixing the actual level count to be correct.


jt-traub:
@ennorehling  this contains your fixes from #279 and the fixes that were needed on top of that to fix the unit tests which then broke.

PR #281, opened 2025-07-24
Fixing errors reported by valgrind

ennorehling:


aregion.cpp
```diff
@@ -791,7 +791,7 @@ Unit *ARegion::GetUnit(int num)
 {
     for(const auto obj : objects) {
         Unit *u = obj->GetUnit(num);
-        if (u) return(u);
+        if (u) return u;
```
jt-traub, line 794: this is a stylistic not functional change, but I approve of it regardless.

object.cpp
```diff
@@ -756,7 +756,7 @@ std::string object_description(int obj)
     int buildable = 1;
     auto pS = FindSkill(o->skill);
     if (o->item == -1 || o->skill == nullptr || !pS || pS->get().flags & SkillType::DISABLED) buildable = 0;
-    if (o->item != I_WOOD_OR_STONE && (ItemDefs[o->item].flags & ItemType::DISABLED)) buildable = 0;
+    if (o->item >= 0 && o->item != I_WOOD_OR_STONE && (ItemDefs[o->item].flags & ItemType::DISABLED)) buildable = 0;
```
jt-traub, line 759: I think the I_WOOD_OR_STONE check needs to come first since it's value is -2.  Or, it could be removed entirely as it's covered under the >= 0 check and thus not necessary here.

aregion.cpp
```diff
@@ -1598,13 +1598,6 @@ int ARegion::ResolveFleetAlias(int alias)
     return f->second;
 }
 
-ARegionList::ARegionList()
```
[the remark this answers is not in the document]
ennorehling: On the contrary, it seems odd to me to have a constructor that does nothing. We have to have a destructor because the internal array, which starts out as null, night have been allocated.

object.cpp
ennorehling: I didn't realize that I_WOOD_OR_STONE was a special case. I dislike these negative constants that signify special cases in variables that are usually an index into an array, Too easy to mess up. Will fix, and clean up the surrounding code a bit.

aregion.cpp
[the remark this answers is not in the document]
jt-traub: Fair, and this (unlike the bracing/etc thing I commented on below) this is one that I wouldn't actually care enough about to change.

object.cpp
```diff
@@ -753,16 +753,22 @@ std::string object_description(int obj)
...
+        buildable = false;
     }
+    else if (o->item >= 0 && (ItemDefs[o->item].flags & ItemType::DISABLED))
+    {
+        buildable = false;
+    }
+    else {
+        buildable = (o->item == I_WOOD_OR_STONE
+            && (ItemDefs[I_WOOD].flags & ItemType::DISABLED)
+            && (ItemDefs[I_STONE].flags & ItemType::DISABLED));
```
jt-traub, line 769: This statement, as opposed to the previous version says an item is only buildable if its item is WOOD_OR_STONE and wood is disabled and stone is disabled.

It's missing a !, or we return as I prefer to the previous case where things are buildable by default and then are not buildable under certain conditions (the condition here being they are not buildable if their are set to WOOD_OR_STONE and both are disabled) :)

jt-traub: This is why the tests are failing btw.

jt-traub:
Subsumed into #282 where I fixed the introduced logic bug.

object.cpp
```diff
@@ -753,16 +753,22 @@ std::string object_description(int obj)
...
-    if (o->item == -1 || o->skill == nullptr || !pS || pS->get().flags & SkillType::DISABLED) buildable = 0;
-    if (o->item >= 0 && o->item != I_WOOD_OR_STONE && (ItemDefs[o->item].flags & ItemType::DISABLED)) buildable = 0;
-    if (
-        o->item == I_WOOD_OR_STONE && (ItemDefs[I_WOOD].flags & ItemType::DISABLED) &&
-        (ItemDefs[I_STONE].flags & ItemType::DISABLED)
-    ) {
-        buildable = 0;
+    if (o->item == -1 || o->skill == nullptr || !pS || pS->get().flags & SkillType::DISABLED)
+    {
+        buildable = false;
```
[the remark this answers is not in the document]
jt-traub: For me 'unneeded' braces are the ones where the if and the thing controlled by the if can be put on the same line.
If they need to be multiple lines, then I open brace at the end of the line with the if and close brace below.

and yes, it's all about vertical space since I like as much code context as possible when reading code :)

PR #290, opened 2025-08-09
Fix the small bug in PR#284 causing news to change

jt-traub:
There is a difference in how struct/class initialization happens, especially when there is no user-provided constructor when an object is construction as
`auto a = new Class` vs `auto a = new Class()`

When the explicit constructor was removed in preference of using the default constructor + member initialization the code was also changed to use the first of those formats instead of the second when constructing a BattleFact
for pvp.   Unfortunately this resulted in the output being
non-deterministic.   Reverting the construction to the
second format fixes this.

Also, re-added and explicit definition of the destructor to the BattleFact class (using = default) since it is required due to the parent class having a pure virtual destructor.  While that didn't contribute to the problem it was technically wrong.

Lastly, cleaned up a couple more uses of const that were redundant.

### DIFF
--- a/battle.cpp
+++ b/battle.cpp
@@ -377,7 +377,7 @@ void AddBattleFact(
 ) {
     if (!Globals->WORLD_EVENTS) return;
 
-    auto fact = new BattleFact;
+    auto fact = new BattleFact();
 
     fact->location = EventLocation::Create(region);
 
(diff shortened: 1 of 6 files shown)


battle.cpp
```diff
@@ -377,7 +377,7 @@ void AddBattleFact(
 ) {
     if (!Globals->WORLD_EVENTS) return;
 
-    auto fact = new BattleFact;
+    auto fact = new BattleFact();
```
jt-traub, line 380: *this* is the bug that was causing the news output to randomize if you had any pvp battles.
While this didn't show up in the snapshot tests, it showed up in the regression suite I run using mid-game NO7 turns.

